PR #8, opened 2023-05-15
Set print with sharing as default for terms

vinciusb:
Sets print with sharing as the default implementation for terms due to the fact that terms can grow exponentially and lead to timeouts and unnecessary time spending.

> Signed-off-by: Vinícius Braga Freire [vinicius.braga.freire@gmail.com](mailto:vinicius.braga.freire@gmail.com)

### DIFF
--- a/carcara/src/ast/printer.rs
+++ b/carcara/src/ast/printer.rs
@@ -321,7 +321,7 @@ impl fmt::Display for Term {
             term_indices: Some(AHashMap::new()),
             term_sharing_variable_prefix: "@p_",
         };
-        printer.write_raw_term(&self).unwrap();
+        printer.write_raw_term(self).unwrap();
         let result = std::str::from_utf8(&buf).unwrap();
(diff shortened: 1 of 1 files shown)


carcara/src/ast/printer.rs
```diff
@@ -315,24 +315,15 @@ fn escape_string(string: &str) -> Cow<str> {
...
-                write!(f, "(lambda {} {})", bindings, term)
-            }
-        }
+        let mut buf = Vec::new();
+        let mut printer = AlethePrinter {
+            inner: &mut buf,
+            term_indices: Some(AHashMap::new()),
+            term_sharing_variable_prefix: "@p_",
+        };
+        printer.write_raw_term(&self).unwrap();
```
bpandreotti, line 324: One very tiny thing: here you don't need to borrow `self` since it is already a `&Term`. You can just do `printer.write_raw_term(self)`. (this was caught by `cargo clippy`)

PR #16, opened 2023-07-11
Set up for Carcará's parallelization

vinciusb:
Sets up the environment and the first features for a concurrent Carcará. In this pull request will be implemented:
- [x] New flags related to the parallel behavior of Carcará.
- [x] Implementation of a new `Pool` struct in a feasible and efficient way to run in parallel.
- [x] Statistics collection and benchmark refactorization.
- [x] Tests adaptation for the new pool architecture.

> Signed-off-by: Vinícius Braga Freire [vinicius.braga.freire@gmail.com](mailto:vinicius.braga.freire@gmail.com)


carcara/src/lib.rs
```diff
@@ -130,38 +135,134 @@ pub enum Error {
     DoesNotReachEmptyClause,
 }
 
-pub fn check<T: io::BufRead>(problem: T, proof: T, options: CarcaraOptions) -> Result<bool, Error> {
-    let (prelude, proof, mut pool) = parser::parse_instance(
+pub fn check<T: io::BufRead>(
```
bpandreotti, line 138: I think it would be better if we still allowed the user to easily check using the single-threaded checker. Ideally, I think there should be two check functions, `check`, that uses the single-threaded checker and has the same behaviour it had before, and `check_parallel`, like you implemented here, which receives a number of threads and stack size. Then in the CLI, if the `--num-threads` option is 1, we use `check`, otherwise we use `check_parallel`

carcara/src/parser/mod.rs
```diff
@@ -511,9 +516,7 @@ impl<'a, R: BufRead> Parser<'a, R> {
                             self.pool
                                 .add(Term::Lambda(BindingList(func_def.params), func_def.body))
                         };
-                        let sort = self
-                            .pool
-                            .add(Term::Sort(self.pool.sort(&lambda_term).clone()));
+                        let sort = self.pool.add(self.pool.sort(&lambda_term).as_ref().clone());
```
bpandreotti, line 519: Same as before
```suggestion
                        let sort = self.pool.sort(&lambda_term);
```

carcara/src/parser/mod.rs
```diff
@@ -805,7 +814,7 @@ impl<'a, R: BufRead> Parser<'a, R> {
             self.next_token()?;
             let var = self.expect_symbol()?;
             let value = self.parse_term()?;
-            let sort = Term::Sort(self.pool.sort(&value).clone());
+            let sort = self.pool.sort(&value).as_ref().clone();
             let sort = self.pool.add(sort);
```
bpandreotti, line 818: Same as before
```suggestion
            let sort = self.pool.sort(&value);
```

carcara/src/parser/mod.rs
```diff
@@ -1018,7 +1027,7 @@ impl<'a, R: BufRead> Parser<'a, R> {
                 p.expect_token(Token::OpenParen)?;
                 let name = p.expect_symbol()?;
                 let value = p.parse_term()?;
-                let sort = p.pool.add(Term::Sort(p.pool.sort(&value).clone()));
+                let sort = p.pool.add(p.pool.sort(&value).as_ref().clone());
```
bpandreotti, line 1030: Same as before
```suggestion
                let sort = p.pool.sort(&value);
```

carcara/src/parser/mod.rs
```diff
@@ -1033,7 +1042,7 @@ impl<'a, R: BufRead> Parser<'a, R> {
             let substitution = bindings
                 .into_iter()
                 .map(|(name, value)| {
-                    let sort = Term::Sort(self.pool.sort(&value).clone());
+                    let sort = self.pool.sort(&value).as_ref().clone();
                     let var = Term::new_var(name, self.pool.add(sort));
```
bpandreotti, line 1046: Same as before
```suggestion
                    let var = Term::new_var(name, self.pool.sort(&value));
```

bpandreotti (changes requested):

carcara/src/checker/parallel/scheduler/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+            if stack.is_empty() {
+                break;
+            }
+            //
+            let AssignedLoad { 0: mut load, 1: load_index } = pq.pop().unwrap();
+            {
+                let top = stack.last().unwrap();
+                let step_weight = get_step_weight(&top.cmds[top.id]);
+                assert!(u64::MAX - step_weight >= load, "Weight balancing overflow!");
+                load += step_weight;
```
bpandreotti, line 160: You can use `checked_add` to protect against overflows:
```suggestion
                load = load
                    .checked_add(step_weight)
                    .expect("Weight balancing overflow!");
```

carcara/src/checker/parallel/scheduler/mod.rs
bpandreotti, line 1: Since the files in this module aren't that large (and are all related) I think we can combine them into just one `checker/parallel/scheduler.rs` file

bpandreotti (changes requested):

carcara/src/checker/parallel/scheduler.rs
```diff
@@ -235,7 +235,7 @@ impl<'a> ScheduleIter<'a> {
     }
 
     /// Returns a slice to the commands of the inner-most open subproof.
-    pub fn current_subproof(&self) -> Option<&[ProofCommand]> {
+    pub fn current_subproof<'s: 'a>(&'s self) -> Option<&[ProofCommand]> {
```
bpandreotti, line 238: Are these lifetime annotations needed?

carcara/src/checker/parallel/mod.rs
```diff
@@ -434,13 +223,127 @@ impl<'c> ParallelProofChecker<'c> {
...
+                    }
+                }
+            }
+            // Verify if any of the other threads found an error and abort in case of positive
+            if should_abort.load(Ordering::Acquire) {
+                break;
+            }
+        }
+
+        // Returns Ok(reached empty clause, isHoley, current thread statistics)
```
bpandreotti, line 332: A small detail: you forgot to update this comment to reflect the new return type
```suggestion
        // Returns Ok(reached empty clause, isHoley)
```

bpandreotti (approved):

PR #22, opened 2023-10-03
feature: add parser support for bit vectors and indexed operators

GuilhermeLLS:


carcara/src/parser/mod.rs
```diff
@@ -1200,6 +1327,18 @@ impl<'a, R: BufRead> Parser<'a, R> {
 
                 Ok(result)
             }
+            Token::OpenParen => {
+                self.next_token()?;
+                // ! problem is here -> we might found double parenthesis. (assume a0 (= (concat #b000 ((_ extract 3 3) #b1000))))
+                if self.current_token == Token::ReservedWord(Reserved::Underscore) {
+                    self.next_token()?;
+                }
```
GuilhermeLLS, line 1335: I had to add this code to prevent the situation where current_token is an underscore. This is an issue, because `parse_indexed_term` assumes the underscore toekn was already consumed. However, I still feel like this is a hack.

carcara/src/ast/pool/mod.rs
```diff
@@ -184,6 +237,54 @@ impl PrimitivePool {
...
+                    IndexedOperator::BvExtract => {
+                        /*
+                        ! extract has restrictions regarding their op_args
+                        ((_ extract i j) (_ BitVec m) (_ BitVec n))
+
+                        where
+                        - i, j, m, n are numerals
+                        - m > i ≥ j ≥ 0,
+                        - n = i - j + 1
+                         */
```
GuilhermeLLS, line 251: @bpandreotti I remember we talked about asserting argument values of indexed operators following SMT-LIB rules.

I did something like this on line [264](https://github.com/ufmg-smite/carcara/pull/22/files#diff-ed7104440530d2d2bc9fc8116dce3419f215a94325ac670b9f99c821072b2a1bR264), but ideally I should return an error.

The problem is that this function only returns an Rc<Term> and I'd have to change it's return type to also include errors, and therefore this would impact other places.

However, this is the smaller of problems. My point is: shouldn't argument values be checked during the parser run? Specifically at the same moment we check the amount of arguments of each Term.

GuilhermeLLS: PS: This is not valid for concat operator specifically, but for every other BV operator. Just used concat as an example.

carcara/src/parser/mod.rs
```diff
@@ -407,6 +410,68 @@ impl<'a, R: BufRead> Parser<'a, R> {
...
+            Operator::BvNot | Operator::BvNeg => {
+                assert_num_args(&args, 1)?;
+                for s in sorts {
+                    let s = s.as_sort().unwrap().clone();
+                    if !matches!(s, Sort::BitVec(_)) {
+                        return Err(ParserError::ExpectedBvSort(s));
+                    }
+                }
+                // TODO: create args value assertion 
+                // assert_args_value(&args, 1..)?;
```
GuilhermeLLS, line 422: Here's the location I meant where I should make args value checking.

https://github.com/ufmg-smite/carcara/pull/22/files#r1394114389

carcara/src/parser/tests.rs
```diff
@@ -661,3 +662,76 @@ fn test_premises_in_subproofs() {
...
+            },
+        ),
+    ];
+
+    for (input, expected_value) in cases {
+        let proof = parse_proof(&mut p, input);
+        assert_eq!(proof.commands.len(), 1);
+        assert_eq!(&proof.commands[0], &expected_value);
+    }
+}
```
GuilhermeLLS, line 737: @bpandreotti any other suggestions for test cases?

carcara/src/parser/error.rs
```diff
@@ -100,6 +123,19 @@ where
...
+    R: Into<Range>,
+{
+    let range = range.into();
+    if range.contains(sequence.len()) {
+        Ok(())
+    } else {
+        Err(ParserError::WrongNumberOfArgs(range, sequence.len()))
+    }
+}
+
```
GuilhermeLLS, line 138: function mentioned to check args value. The implementation is not there yet, I just copied the body from `assert_num_args`

GuilhermeLLS:
I'll remove all the dbg! and print from the code, just want to make sure other things are good to go first.

carcara/src/ast/printer.rs
```diff
@@ -117,13 +123,13 @@ impl<'a> PrintProof for AlethePrinter<'a> {
         while let Some(command) = iter.next() {
             match command {
                 ProofCommand::Assume { id, term } => {
-                    write!(self.inner, "(assume {} ", quote_symbol(id))?;
+                    write!(self.inner, "(assume {} ", id)?;
```
bpandreotti, line 126: You accidentally reverted commit b9cf66b4b664b955e8dbd4fe4ae1b1d9fe736b48. Probably an error when rebasing.

carcara/src/ast/printer.rs
```diff
@@ -408,6 +406,7 @@ impl fmt::Display for Sort {
             Sort::String => write!(f, "String"),
             Sort::RegLan => write!(f, "RegLan"),
             Sort::Array(x, y) => write_s_expr(f, "Array", &[x, y]),
+            Sort::BitVec(w) => write!(f, "BitVec, {}",w),
```
bpandreotti, line 409: This should be:
```suggestion
            Sort::BitVec(w) => write!(f, "(_ BitVec {})",w),
```

carcara/src/ast/printer.rs
```diff
@@ -360,6 +357,7 @@ impl fmt::Display for Constant {
                 }
             }
             Constant::String(s) => write!(f, "\"{}\"", escape_string(s)),
+            Constant::BitVec(val, width) => write!(f, "{}, {}", val, width), // TODO: comeback to this
```
bpandreotti, line 360: There are multiple valid ways of printing bitvector constants, but I think the simplest way is to use the `(_ bv...)` syntax:
```suggestion
            Constant::BitVec(val, width) => write!(f, "(_ bv{} {})", val, width),
```

carcara/src/parser/error.rs
```diff
@@ -85,6 +104,10 @@ pub enum ParserError {
     /// The parser encountered the end of the input while it was still inside a subproof.
     #[error("subproof '{0}' was not closed")]
     UnclosedSubproof(String),
+
+    /// The parser could not get bitvector width from Sort
+    #[error("could not get bitvector width from sort '{0}'")]
+    UnreachableBitVecWidth(Sort),
```
bpandreotti, line 110: This error is not used anywhere, I think it can be removed.

carcara/src/parser/tests.rs
bpandreotti: Looks good to me

carcara/src/ast/pool/mod.rs
bpandreotti: > However, this is the smaller of problems. My point is: shouldn't argument values be checked during the parser run? Specifically at the same moment we check the amount of arguments of each Term.

I probably misspoke before, but you're absolutely right, these restrictions on the arguments should be checked when parsing the term (in `Parser::make_indexed_op` to be precise). Here, we can just assume that the term is well-sorted, that is, that the arguments follow the restrictions.

carcara/src/ast/pool/mod.rs
```diff
@@ -184,6 +237,54 @@ impl PrimitivePool {
...
+                        else {
+                            unreachable!()
+                        };
+                        if !(m > i && i >= j && j >= Integer::ZERO) {
+                            panic!("Wrong arguments for extract") // todo create error for this
+                        }
+                        Sort::BitVec(i - j + Integer::ONE)
+                    }
+                    IndexedOperator::ZeroExtend | IndexedOperator::SignExtend => {
+                        let Constant::Integer(extension_width) = op_args[0].clone() else {
```
bpandreotti, line 270: By the way, it might be nice to add some helper methods like `Constant::as_integer` (that returns an `Option<Integer>`) so you can just call `op_args[0].as_integer().unwrap()`. Something similar to the `Term::as_{var,sort,etc}` methods.

carcara/src/parser/mod.rs
```diff
@@ -407,6 +410,68 @@ impl<'a, R: BufRead> Parser<'a, R> {
...
+            | Operator::BvAnd
+            | Operator::BvOr
+            | Operator::BvXor => {
+                assert_num_args(&args, 2..)?;
+                for s in sorts {
+                    let s = s.as_sort().unwrap().clone();
+                    if !matches!(s, Sort::BitVec(_)) {
+                        return Err(ParserError::ExpectedBvSort(s));
+                    }
+                }
```
bpandreotti, line 443: Here, besides making sure that the terms are all bitvectors, we also have to make sure they have the same size (e.g., `(bvadd #b01 #b001)` should be rejected because the arguments have different widths). I think a nice way to do this is to first check that the first argument is a bitvector, and then checking that all arguments have the same sort.
```suggestion
                assert_num_args(&args, 2..)?;
                let first_sort = sorts[0].as_sort().unwrap().clone();
                if !matches!(first_sort, Sort::BitVec(_)) {
                    return Err(ParserError::ExpectedBvSort(first_sort));
                }
                SortError::assert_all_eq(
                    &sorts
                        .iter()
                        .map(|op| op.as_sort().unwrap())
                        .collect::<Vec<&Sort>>(),
                )?;
```

carcara/src/parser/mod.rs
```diff
@@ -407,6 +410,68 @@ impl<'a, R: BufRead> Parser<'a, R> {
...
+            | Operator::BvSGt
+            | Operator::BvSGe => {
+                dbg!(&op, &args);
+                assert_num_args(&args, 2)?;
+                for s in sorts {
+                    let s = s.as_sort().unwrap().clone();
+                    if !matches!(s, Sort::BitVec(_)) {
+                        return Err(ParserError::ExpectedBvSort(s));
+                    }
+                }
```
bpandreotti, line 473: Same as above, the arguments should have the same width

carcara/src/parser/mod.rs
GuilhermeLLS:
```rust
SortError::assert_all_eq(
                    &sorts
                        .iter()
                        .map(|op| op.as_sort().unwrap())
                        .collect::<Vec<&Sort>>(),
                )?;
```

is this code already enough to check if arguments have the same width?

GuilhermeLLS: and also, I don't think all arguments should have the same width for `Concat`. (Double checking w/ you)

GuilhermeLLS: one more thing: besides the check provided from the code above, do I also need to check the values of the values of the arguments for non-indexed operators?


something like `assert_args_value`

bpandreotti: > ```rust
> SortError::assert_all_eq(
>                     &sorts
>                         .iter()
>                         .map(|op| op.as_sort().unwrap())
>                         .collect::<Vec<&Sort>>(),
>                 )?;
> ```
> 
> is this code enough to check if arguments have the same width?

Yeah!

bpandreotti: > and also, I don't think all arguments should have the same width for `Concat`. (Double checking w/ you)

You're right, for string concatenation the arguments can have different widths (but they should all be bitvectors)

bpandreotti: > one more thing: besides the check provided from the code above, do I also need to check the values of the values of the arguments for non-indexed operators?
> 
> something like `assert_args_value`

No. You only have to check that for indexed operators, since they have `op_args` that need to follow certain restrictions.

carcara/src/parser/error.rs
```diff
@@ -81,8 +81,12 @@ pub enum ParserError {
     #[error("expected {0} arguments, got {1}")]
     WrongNumberOfArgs(Range, usize),
 
+    /// The argument values are not in the expected range.
     #[error("expected argument value to be greater than {0}, got {1}")]
-    ExpectedGreaterThanArgs(usize, usize),
+    ExpectedGreaterThanArgs(Range, usize),
```
bpandreotti, line 86: Since this now accepts any `Range`, I think the name is a bit confusing. It would be clearer if it were something like
```suggestion
    #[error("expected argument value to be {0}, got {1}")]
    WrongValueOfArgs(Range, usize),
```

carcara/src/checker/mod.rs
```diff
@@ -555,6 +555,7 @@ impl<'c> ProofChecker<'c> {
             "la_mult_pos" => extras::la_mult_pos,
             "la_mult_neg" => extras::la_mult_neg,
             "mod_simplify" => extras::mod_simplify,
+            "bitblast_extract" => bvextract::extract,
```
bpandreotti, line 558: We should keep only the parser changes in this PR. The implementation of the actual rules should be in a different PR.

GuilhermeLLS: ooops.

carcara/src/parser/mod.rs
```diff
@@ -1222,6 +1319,89 @@ impl<'a, R: BufRead> Parser<'a, R> {
...
+                assert_num_args(&op_args, 2)?;
+                assert_num_args(&args, 0)?;
+                let Constant::Integer(value) = op_args[0].clone() else {
+                    return Err(ParserError::ExpectedIntegerConstant(op_args[0].clone()));
+                };
+                let Constant::Integer(width) = op_args[1].clone() else {
+                    return Err(ParserError::ExpectedIntegerConstant(op_args[1].clone()));
+                };
+                assert_indexed_op_args_value(&[op_args[0].clone()], 0..)?;
+                assert_indexed_op_args_value(&[op_args[1].clone()], 1..)?;
```
GuilhermeLLS, line 1354: had to update this to allow constants like `(_ bv0 4)` to be correctly parsed.
@bpandreotti

GuilhermeLLS:
can you help me fixing the 2 tests that are failing on the CI? I'm having a hard time with them 😓  @bpandreotti

carcara/src/parser/mod.rs
```diff
@@ -1290,6 +1475,14 @@ impl<'a, R: BufRead> Parser<'a, R> {
 
                 Ok(result)
             }
+            Token::OpenParen => {
+                self.next_token()?;
+                self.expect_token(Token::ReservedWord(Reserved::Underscore))?;
+                let (op, op_args) = self.parse_indexed_operator()?;
+                let args = self.parse_sequence(Self::parse_term, true)?;
+                self.make_indexed_op(op, op_args, args)
+                    .map_err(|err| Error::Parser(err, head_pos))
```
bpandreotti, line 1484: The tests are failing because, when adding support for parsing indexed operations (that start with the tokens `((`), we accidentally broke the parsing of `lambda` terms. This is a bit tricky to solve, but I think the cleanest way to do this without adding lookahead is this:
```suggestion
                if self.current_token == Token::ReservedWord(Reserved::Underscore) {
                    self.next_token()?;
                    let (op, op_args) = self.parse_indexed_operator()?;
                    let args = self.parse_sequence(Self::parse_term, true)?;
                    self.make_indexed_op(op, op_args, args)
                        .map_err(|err| Error::Parser(err, head_pos))
                } else {
                    let func = self.parse_application()?;
                    let args = self.parse_sequence(Self::parse_term, true)?;
                    self.make_app(func, args)
                        .map_err(|err| Error::Parser(err, head_pos))
                }
```

bpandreotti:
As for the formatting, you can run `rustfmt` on the whole project with `rustfmt $(find ./carcara/src -name '*.rs')`, and it should fix the CI errors. Though in the future it might be useful to configure your editor to format on save.

carcara/src/parser/mod.rs
GuilhermeLLS: ty

bpandreotti (approved):

PR #23, opened 2023-10-27
Adds support for Strings and Regular Expressions operators in AST

vinisilvag:


carcara/src/ast/mod.rs
```diff
@@ -272,6 +272,107 @@ pub enum Operator {
...
+    StrConcat,
+
+    /// The `str.len` operator.
+    StrLen,
+
+    /// The `str.<` operator.
+    LexOrdering,
+
+    /// The `str.<=` operator.
+    ReflexiveClosure,
```
bpandreotti, line 287: I think it would be clearer to rename these to `StrLessThan` and `StrLessEq`

bpandreotti (changes requested):

bpandreotti (approved):

PR #24, opened 2023-11-23
feat: add bitblast_extract rule

GuilhermeLLS:


carcara/src/checker/rules/bvextract.rs
```diff
@@ -0,0 +1,58 @@
...
+    match right.as_ref() {
+        Term::Op(Operator::BvBbTerm, args) => {
+            assert_eq!(args.len(), size);
+            let (_, left_x_args) = left_x.as_op().unwrap();
+            // dbg!(args, left_x_args);
+            assert_eq!(args, &left_x_args[left_j..=left_i]);
+            Ok(())
+        }
+        _ => Err(CheckerError::Extract(ExtractError::WrongRightTerm)),
+    }
```
GuilhermeLLS, line 39: I have a question:

the clause below is valid, and therefore, is validated by the parser and the rule.
```
(step t2 (cl (= ((_ extract 1 0) (bbterm ((_ bit_of 0) (ite a #b110 #b011)) ((_ bit_of 1) (ite a #b110 #b011)) ((_ bit_of 2) (ite a #b110 #b011)))) (bbterm ((_ bit_of 0) (ite a #b110 #b011)) ((_ bit_of 1) (ite a #b110 #b011))))) :rule bitblast_extract)
```

the clause below is not semantically invalid from the parser perspective, but the parser is still able to know that it has a bit_of missing. (run the tests to see the error)

```
(step t3 (cl (= ((_ extract 1 0) (bbterm ((_ bit_of 0) (ite a #b110 #b011)) ((_ bit_of 1) (ite a #b110 #b011)) ((_ bit_of 2) (ite a #b110 #b011)))) (bbterm ((_ bit_of 0) (ite a #b110 #b011))))) :rule bitblast_extract)
```

Should the parser really be able to know this? Cause I feel like this is something related to the rule itself. Or did I just pick a bad example?

HanielB: What is the error that is given by the parser?

GuilhermeLLS:
```
parser error: sort error: expected '(_ BitVec 2)', got '(_ BitVec 1)' (on line 1, column 15)
```

this error makes sense, but I thought this should be a responsibility of the rule, not the parser.

carcara/src/checker/rules/bvextract.rs
bpandreotti, line 1: Instead of having a file for each bitvector rule, I think it makes more sense to have a single file for all bitvector rules (something like `src/checker/rules/bitvectors.rs`).

HanielB:
Maybe it'd be best to split the addition of the rule checkers into different PRs

GuilhermeLLS:
> Maybe it'd be best to split the addition of the rule checkers into different PRs

Why? 🤔

HanielB:
So each one is simpler and easier to review. :)

HanielB:
Thanks! Also please update this one to remove the code added in the other PRs.

bpandreotti (approved):

PR #30, opened 2024-01-17
Adds the concat_eq rule

vinisilvag:


carcara/src/checker/rules/strings.rs
```diff
@@ -6,25 +6,31 @@ use super::{
...
+                        s.chars()
+                            .map(|c| pool.add(Term::Const(Constant::String(c.to_string())))),
+                    );
+                }
+                Term::Op(Operator::StrConcat, _) => {
+                    flattened.extend(flatten(arg.clone(), pool));
+                }
+                _ => {
+                    flattened.push(arg.clone());
+                }
```
bpandreotti, line 30: Notice that what you're doing in this `match` expression is basically what the body of `flatten` already does. So to simplify you can just call it directly:
```suggestion
            flattened.extend(flatten(arg.clone(), pool));
```

carcara/src/ast/mod.rs
```diff
@@ -739,6 +739,16 @@ impl Term {
...
 
+    /// Tries to extract a `bool` from a term. Returns `Some` if the term is an boolean
+    /// constant.
+    pub fn as_bool(&self) -> Option<bool> {
+        if let Term::Var(name, sort) = self {
+            Some(sort.as_sort() == Some(&Sort::Bool) && name == "true")
+        } else {
+            None
+        }
+    }
```
bpandreotti, line 750: This implementation is wrong. Any variable whose sort is not `Bool` will return `Some(false)`, implying that it is the boolean constant `false`. Instead, if the sort is not `Bool`, or the variable name is not `true` or `false`, you should return `None`:
```suggestion
    pub fn as_bool(&self) -> Option<bool> {
        match self {
            Term::Var(name, sort) if sort.as_sort() == Some(&Sort::Bool) => match name.as_str() {
                "true" => Some(true),
                "false" => Some(false),
                _ => None,
            },
            _ => None,
        }
    }
```

carcara/src/checker/rules/strings.rs
```diff
@@ -0,0 +1,136 @@
...
+        Operator::Equals,
+        vec![conc_left_concat, conc_right_concat],
+    ));
+
+    assert_eq(&expected, &expanded_conc)
+}
+
+mod tests {
+    #[test]
+    fn concat_eq() {
```
bpandreotti, line 97: Overall the tests are pretty comprehensive, but could you add a few tests for invalid cases? For example, a test case where the prefixes don't match, or the argument given was wrong, etc. It's arguably more important to test for these cases because real-world proofs already test for the valid cases (and also, it's better to reject a valid proof than to accept an invalid one).

vinisilvag: Sure, I'll do that!

PR #33, opened 2024-03-19
[rules] [strings] Adds the `concat_conflict` rule and unit tests

vinisilvag:
This PR adds the `concat_conflict` rule from the Strings theory.

Rule reference can be found in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/proofs/proof_rules.html#_CPPv4N4cvc59ProofRule15CONCAT_CONFLICTE) and in the [AletheLF specification](https://github.com/cvc5/cvc5/blob/3d249fa410d350d418e68f893fe175f80ab9ce10/proofs/alf/cvc5/rules/Strings.smt3#L67-L94).

### DIFF
--- a/carcara/src/checker/rules/strings.rs
+++ b/carcara/src/checker/rules/strings.rs
@@ -25,7 +25,7 @@ fn flatten(term: Rc<Term>, pool: &mut dyn TermPool) -> Vec<Rc<Term>> {
 
 fn reconstruct_term(terms: Vec<Rc<Term>>, pool: &mut dyn TermPool) -> Rc<Term> {
     match terms.len() {
-        0 => pool.add(Term::Const(Constant::String(String::from("")))),
+        0 => pool.add(Term::Const(Constant::String("".to_owned()))),
         1 => terms[0].clone(),
         _ => pool.add(Term::Op(Operator::StrConcat, terms)),
     }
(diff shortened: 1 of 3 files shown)


carcara/src/checker/rules/strings.rs
```diff
@@ -57,6 +57,52 @@ fn is_prefix(
...
+        s_flat.reverse();
+        t_flat.reverse();
+    }
+
+    if s_flat.is_empty() {
+        return Err(CheckerError::ExpectedConcatApplication(s));
+    }
+    if t_flat.is_empty() {
+        return Err(CheckerError::ExpectedConcatApplication(t));
+    }
```
bpandreotti, line 82: Instead of creating a new error variant, you can use `TermOfWrongForm` (this is the error variant returned by `match_term_err!`)
```suggestion
    if s_flat.is_empty() {
        return Err(CheckerError::TermOfWrongForm("(str.++ ...)", s));
    }
    if t_flat.is_empty() {
        return Err(CheckerError::TermOfWrongForm("(str.++ ...)", t));
    }
```

carcara/src/checker/rules/strings.rs
```diff
@@ -161,6 +188,50 @@ pub fn concat_unify(
...
+    assert_num_args(args, 1)?;
+    assert_clause_len(conclusion, 1)?;
+
+    let term = get_premise_term(&premises[0])?;
+    let rev = args[0].as_term()?.as_bool_err()?;
+
+    if conclusion[0].as_bool_err()? {
+        return Err(CheckerError::ExpectedFalseBoolConstant(
+            conclusion[0].clone(),
+        ));
```
bpandreotti, line 211: There already is an error variant that covers this case, `ExpectedBoolConstant`:
```suggestion
        return Err(CheckerError::ExpectedBoolConstant(
            false,
            conclusion[0].clone(),
        ));
```

bpandreotti (approved):

PR #35, opened 2024-05-07
[rules] [strings] Add support for `concat_split` rule and splitted rules in prefix and suffix version

vinisilvag:
This PR adds the `concat_csplit` rule from the Strings theory.
Rule reference can be found in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/proofs/proof_rules.html#_CPPv4N4cvc59ProofRule12CONCAT_SPLITE) and in the [AletheLF specification](https://github.com/cvc5/cvc5/blob/14eae04ec806140aa8e92293b69b2ca7a251d895/proofs/alf/cvc5/rules/Strings.smt3#L68-L95).

Furthermore, it:
* Splits the rule into two versions: application over prefixes and over suffixes;
* Makes the necessary modifications to allow other rules to be split in this way;
* Standardizes variable names to how they appear in AletheLF documentation;
* Fixes a bug in the function that creates Skolems in the `concat_csplit` rule.

### DIFF
--- a/carcara/src/checker/rules/strings.rs
+++ b/carcara/src/checker/rules/strings.rs
@@ -810,74 +810,74 @@ mod tests {
             "Simple working examples" {
                 r#"(assume h1 (= (str.++ "a" "b" b) (str.++ "a" c)))
                    (assume h2 (not (= (str.len "a") 0)))
-                   (step t1 (cl (= "a" (str.++ "a" (str.substr "a" 0 (- (str.len "a") 1))))) :rule concat_csplit :premises (h1 h2) :args (false))"#: true,
+                   (step t1 (cl (= "a" (str.++ "a" (str.substr "a" 1 (- (str.len "a") 1))))) :rule concat_csplit :premises (h1 h2) :args (false))"#: true,
                 r#"(assume h1 (= (str.++ a "b" c) (str.++ "ab" c)))
                    (assume h2 (not (= (str.len (str.++ a "b")) 0)))
-                   (step t1 (cl (= (str.++ a "b") (str.++ "a" (str.substr (str.++ a "b") 0 (- (str.len (str.++ a "b")) 1))))) :rule concat_csplit :premises (h1 h2) :args (false))"#: true,
+                   (step t1 (cl (= (str.++ a "b") (str.++ "a" (str.substr (str.++ a "b") 1 (- (str.len (str.++ a "b")) 1))))) :rule concat_csplit :premises (h1 h2) :args (false))"#: true,
                 r#"(assume h1 (= (str.++ d (str.++ c (str.++ b "a"))) (str.++ "b" (str.++ "c" d))))
                    (assume h2 (not (= (str.len (str.++ d c)) 0)))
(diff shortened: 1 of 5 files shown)


carcara/src/checker/rules/strings.rs
```diff
@@ -15,91 +15,162 @@ fn flatten(term: Rc<Term>, pool: &mut dyn TermPool) -> Vec<Rc<Term>> {
...
+    pref: Rc<Term>,
+    polyeq_time: &mut Duration,
+) -> Result<Vec<Rc<Term>>, CheckerError> {
+    is_prefix_or_suffix(pool, term, pref, true, polyeq_time)
+}
+
 type Suffixes = (Vec<Rc<Term>>, Vec<Rc<Term>>);
 
-fn strip_prefix(
+fn strip_prefix_or_suffix(
```
bpandreotti, line 143: It would be nice to have a comment documenting what this functions does. This is especially true for these helper functions that are used in multiple rules (e.g. `strip_prefix_or_suffix`, `flatten`, `expand_string_constants`, etc.).

carcara/src/checker/rules/strings.rs
```diff
@@ -15,91 +15,162 @@ fn flatten(term: Rc<Term>, pool: &mut dyn TermPool) -> Vec<Rc<Term>> {
...
+                .collect();
+            let new_args = args
+                .iter()
+                .map(|term| expand_string_constants(pool, term))
+                .collect();
+            pool.add(Term::ParamOp {
+                op: *op,
+                op_args: new_op_args,
+                args: new_args,
+            })
```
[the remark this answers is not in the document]
vinisilvag: Oops, I hadn't seen that in the documentation. Already fixed!

carcara/src/checker/rules/strings.rs
```diff
@@ -6,6 +6,19 @@ use crate::{ast::*, checker::error::CheckerError};
 use rug::Integer;
 use std::{cmp, time::Duration};
 
+/// A function that takes an `Rc<Term>` and returns a vector corresponding to
+/// the flat form of that term.
+///
+/// In this flat form, all `str.++` applications are dissolved and every
+/// argument of those applications is inserted into the vector (e.g., `(str.++
+/// a (str.++ b c))` would lead to `[a, b, c]`, where `a`, `b` e `c` são
+/// arbitrary String terms). Furthermore, all String constants are broken
```
bpandreotti, line 15:
```suggestion
/// a (str.++ b c))` would lead to `[a, b, c]`, where `a`, `b` and `c` are
/// arbitrary String terms). Furthermore, all String constants are broken
```

carcara/src/checker/rules/strings.rs
```diff
@@ -6,6 +6,19 @@ use crate::{ast::*, checker::error::CheckerError};
...
+/// into constants of size one and are inserted into the vector too.
+///
+/// All String terms that aren't `str.++` applications can be seen as `(str.++ term "")`.
+/// So, applying `flatten` to them would lead to `[term]`. Furthermore,
+/// the empty String isn't mapped to any entry in the vector, so `flatten`
+/// applied to `""` lead to an empty vector.
```
bpandreotti, line 21:
```suggestion
/// applied to `""` leads to an empty vector.
```

PR #37, opened 2024-05-10
Update parser for new syntax of anchors

HanielB:


carcara/src/parser/mod.rs
```diff
@@ -1011,14 +1011,13 @@ impl<'a, R: BufRead> Parser<'a, R> {
...
     /// Parses an argument for an `anchor` proof command. This can be either a variable binding of
-    /// the form `(<symbol> <sort>)` or an assignment, of the form `(:= <symbol> <term>)`.
+    /// the form `(<symbol> <sort>)` or an assignment, of the form `(:= (<symbol> <sort>) <term>)`.
     fn parse_anchor_argument(&mut self) -> CarcaraResult<AnchorArg> {
         self.expect_token(Token::OpenParen)?;
         Ok(if self.current_token == Token::Keyword("=".into()) {
             self.next_token()?;
-            let var = self.expect_symbol()?;
+            let (var, sort) = self.parse_sorted_var()?;
             let value = self.parse_term()?;
```
bpandreotti, line 1020: We are doing nothing here to prevent the value sort from being different from the given sort, e.g. `(:= (x Int) false)`. One way to solve this would be to use `parse_term_expecting_sort`:
```suggestion
            let value = self.parse_term_expecting_sort(sort.as_sort().unwrap())?;
```

PR #42, opened 2024-06-12
[rules] [strings] Add support for `concat_cprop` rule

vinisilvag:
This PR adds the `concat_cprop` rule from the Strings theory.
Rule reference can be found in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/proofs/proof_rules.html#_CPPv4N4cvc59ProofRule12CONCAT_CPROPE) and in the [AletheLF specification](https://github.com/cvc5/cvc5/blob/942f4bac8da99fa84d2629f076dce64908026d6e/proofs/alf/cvc5/rules/Strings.smt3#L122-L143).

### DIFF
--- a/carcara/src/checker/error.rs
+++ b/carcara/src/checker/error.rs
@@ -141,6 +141,9 @@ pub enum CheckerError {
     #[error("expected term {0} to be a suffix of {1}")]
     ExpectedToBeSuffix(Rc<Term>, Rc<Term>),
 
+    #[error("expected term {0} to not be empty")]
+    ExpectedToNotBeEmpty(Rc<Term>),
+
     #[error("this rule can only be used in the last step of a subproof")]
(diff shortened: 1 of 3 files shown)


HanielB:
From discussing offline: 
- factor out the structural checks that are identical in both the prefix and suffix case
- do not consider a Carcara failure (i.e., `unreachable`) the ill-formed case in `overlap`, but rather add checks to guarantee that the rule input will lead to a successfull application or lead to a `CheckerError`

vinisilvag:
> From discussing offline:
> 
> * factor out the structural checks that are identical in both the prefix and suffix case
> * do not consider a Carcara failure (i.e., `unreachable`) the ill-formed case in `overlap`, but rather add checks to guarantee that the rule input will lead to a successfull application or lead to a `CheckerError`

PR #44, opened 2024-06-28
[rules] [strings] Add support for the String length core rules and extend `PolyeqComparator` for String comparisons

vinisilvag:
This PR adds the `string_decompose`, `string_length_pos` and `string_length_non_empty` rules from the Strings theory.
Rules reference can be found in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/proofs/proof_rules.html#_CPPv4N4cvc59ProofRule16STRING_DECOMPOSEE) and in the [AletheLF specification](https://github.com/cvc5/cvc5/blob/994c86b5dc4cc6d3a54f9fc1cdda593fd9372012/proofs/alf/cvc5/rules/Strings.smt3#L7-L17).

Furthermore, it extends the `PolyeqComparator` to properly reason over String concatenations and String constants. This extension allows semantically equal Strings, such as `(str.++ "a" "b")` and `"ab"`, to be identified as equal by the comparator.

### DIFF
--- a/carcara/src/ast/polyeq.rs
+++ b/carcara/src/ast/polyeq.rs
@@ -7,13 +7,20 @@
 //! - `alpha_equiv` compares terms by alpha-equivalence, meaning it implements equality of terms
 //! modulo renaming of bound variables.
 
+use rug::Rational;
+
 use super::{
     AnchorArg, BindingList, Constant, Operator, ProofArg, ProofCommand, ProofStep, Rc, Sort,
     Subproof, Term,
 };
 use crate::utils::HashMapStack;
-use rug::Rational;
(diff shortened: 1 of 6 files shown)


carcara/src/ast/polyeq.rs
bpandreotti, line 1: I think it would be best to add a parameter to `PolyeqComparator` to control this new string comparison behaviour. Something similar to what is currently done for `is_mod_reordering`/`is_alpha_equivalence`/`is_mod_nary`.

vinisilvag: Sure, I'll do that!

PR #46, opened 2024-10-03
[rules] [strings] Add support for some core Strings theory rules that reason about Regular expressions

vinisilvag:
This PR adds the `re_inter`, `re_unfold_neg` and `re_unfold_neg_concat_fixed` rules from the Strings theory.
For simplification reasons, the `re_unfold_neg_concat_fixed` rule was splitted in a prefix version and suffix version (resp. `re_unfold_neg_concat_fixed_prefix` and `re_unfold_neg_concat_fixed_suffix`), just like it was done for other previously implemented rules.

Future PR's will add the `re_unfold_pos` rule.

Rules reference can be found in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/api/cpp/enums/proofrule.html#_CPPv4N4cvc59ProofRule8RE_INTERE) and in the [CPC specification](https://github.com/cvc5/cvc5/blob/973ec555d552c7e0d0747308f18498f9838327e2/proofs/eo/cpc/rules/Strings.eo#L314-L416).

### DIFF
--- a/carcara/src/checker/rules/strings.rs
+++ b/carcara/src/checker/rules/strings.rs
@@ -341,11 +341,7 @@ fn str_fixed_len_re(pool: &mut dyn TermPool, r: Rc<Term>) -> Result<usize, Check
                 let r_2_concat = concat(pool, r_2.to_vec());
                 Ok(str_fixed_len_re(pool, r_1.clone())? + str_fixed_len_re(pool, r_2_concat)?)
             } else {
-                Err(CheckerError::WrongNumberOfTermsInOp(
-                    Operator::ReConcat,
-                    (2..).into(),
-                    args.len(),
-                ))
+                unreachable!()
             }
         }
         Term::Op(Operator::ReAllChar, _) => Ok(1),
@@ -357,11 +353,7 @@ fn str_fixed_len_re(pool: &mut dyn TermPool, r: Rc<Term>) -> Result<usize, Check
                     _ => Err(CheckerError::LengthCannotBeEvaluated(r.clone())),
                 }
             } else {
-                Err(CheckerError::WrongNumberOfTermsInOp(
-                    Operator::StrToRe,
-                    (1).into(),
-                    args.len(),
-                ))
+                unreachable!()
             }
         }
         Term::Op(Operator::ReUnion, args) => {
@@ -381,11 +373,7 @@ fn str_fixed_len_re(pool: &mut dyn TermPool, r: Rc<Term>) -> Result<usize, Check
                     }
                 }
             } else {
-                Err(CheckerError::WrongNumberOfTermsInOp(
-                    Operator::ReUnion,
-                    (2..).into(),
-                    args.len(),
-                ))
+                unreachable!()
             }
         }
         Term::Op(Operator::ReIntersection, args) => {
@@ -405,11 +393,7 @@ fn str_fixed_len_re(pool: &mut dyn TermPool, r: Rc<Term>) -> Result<usize, Check
                     }
                 }
             } else {
-                Err(CheckerError::WrongNumberOfTermsInOp(
-                    Operator::ReIntersection,
-                    (2..).into(),
-                    args.len(),
-                ))
+                unreachable!()
             }
         }
         _ => Err(CheckerError::LengthCannotBeEvaluated(r.clone())),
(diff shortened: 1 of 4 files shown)


carcara/src/checker/rules/strings.rs
```diff
@@ -323,6 +327,79 @@ fn extract_arguments(t: &Rc<Term>) -> Result<Vec<Rc<Term>>, CheckerError> {
...
 }
 
+fn singleton_elim(pool: &mut dyn TermPool, r_list: Vec<Rc<Term>>) -> Rc<Term> {
+    match r_list.len() {
+        1 => r_list[0].clone(),
+        _ => pool.add(Term::Op(Operator::ReConcat, r_list)),
+    }
+}
+
+fn str_fixed_len_re(pool: &mut dyn TermPool, r: Rc<Term>) -> Result<usize, CheckerError> {
```
bpandreotti, line 337: It would be nice to add a comment explaining what this function does -- it is not very clear from just the name.

carcara/src/checker/rules/strings.rs
```diff
@@ -323,6 +327,79 @@ fn extract_arguments(t: &Rc<Term>) -> Result<Vec<Rc<Term>>, CheckerError> {
...
+                    let new_r_2 = pool.add(Term::Op(Operator::ReIntersection, r_2.to_vec()));
+                    if str_fixed_len_re(pool, new_r_2)? == n {
+                        Ok(n)
+                    } else {
+                        Err(CheckerError::LengthCannotBeEvaluated(r.clone()))
+                    }
+                }
+            } else {
+                unreachable!()
+            }
```
bpandreotti, line 397: Similarly to the `ReConcat` case, in these two cases you are constructing a `re.union`/`re.inter` term with the remaining arguments just to call `str_fixed_re_len` recursively, which is not very efficient (for each argument, you allocate a vector with all the remaining arguments). I think it would be best to call `str_fixed_re_len` on each argument directly to get their lenghts, and then check if they are all the same.

Also, since these two cases are almost identical, differing only in which terms they ignore, it might make sense to put this logic in a separate function and call it in each case.

bpandreotti (changes requested):

carcara/src/checker/rules/strings.rs
```diff
@@ -323,6 +327,79 @@ fn extract_arguments(t: &Rc<Term>) -> Result<Vec<Rc<Term>>, CheckerError> {
...
+        Term::Op(Operator::ReRange, _) => Ok(1),
+        Term::Op(Operator::StrToRe, args) => {
+            if let Some(s_1) = args.first() {
+                match s_1.as_ref() {
+                    Term::Const(Constant::String(s)) => Ok(s.len()),
+                    _ => Err(CheckerError::LengthCannotBeEvaluated(r.clone())),
+                }
+            } else {
+                unreachable!()
+            }
```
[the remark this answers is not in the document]
vinisilvag: Sure!

carcara/src/checker/rules/strings.rs
```diff
@@ -323,6 +327,79 @@ fn extract_arguments(t: &Rc<Term>) -> Result<Vec<Rc<Term>>, CheckerError> {
...
+
+fn str_fixed_len_re(pool: &mut dyn TermPool, r: Rc<Term>) -> Result<usize, CheckerError> {
+    match r.as_ref() {
+        Term::Op(Operator::ReConcat, args) => {
+            if let [r_1, r_2 @ ..] = &args[..] {
+                let r_2_concat = concat(pool, r_2.to_vec());
+                Ok(str_fixed_len_re(pool, r_1.clone())? + str_fixed_len_re(pool, r_2_concat)?)
+            } else {
+                unreachable!()
+            }
```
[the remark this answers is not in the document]
vinisilvag: It is not correct, indeed :(

I confused this function with another one I previously implemented and ended up calling the wrong one. Also, I did not call `str_fixed_len_re` for each argument directly because I got too focused on how CPC represents `re.++` and `str.++` terms (`(str.++ a b c)` is transformed into `(str.++ a (str.++ b (str.++ c "")))`). However, with the representation in arrays that we have internally, it is, in fact, the same as applying `str_fixed_len_re` directly to each argument.

I am going to change that there and in other places where a similar behavior occurs.

carcara/src/checker/rules/strings.rs
vinisilvag: Done!

bpandreotti (approved):

PR #47, opened 2024-10-13
Update for arguments being only terms

HanielB:
The main impact is that ProofArg is removed and everything is made more uniform
since arguments are only terms.

### DIFF
--- a/carcara/src/checker/error.rs
+++ b/carcara/src/checker/error.rs
@@ -219,9 +219,6 @@ pub enum CongruenceError {
 /// Errors relevant to the rules dealing with quantifiers.
 #[derive(Debug, Error)]
 pub enum QuantifierError {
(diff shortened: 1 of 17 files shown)


carcara/src/checker/error.rs
```diff
@@ -225,6 +219,9 @@ pub enum CongruenceError {
 /// Errors relevant to the rules dealing with quantifiers.
 #[derive(Debug, Error)]
 pub enum QuantifierError {
+    #[error("lhs of argument equality is not a variable: '{0}'")]
+    ArgNotVar(Rc<Term>),
```
bpandreotti, line 223: Why was this error type added? It does not seem to be used anywhere

HanielB: It was used in an intermediate version. I had forgotten to delete it. It's removed now.

bpandreotti (approved):

PR #52, opened 2025-01-01
Fix small bug in checking of connective_def if the connective is forall

Lachnitt:
Switch forall and exists


bpandreotti (approved):

carcara/src/checker/rules/tautology.rs
bpandreotti, line 1: We also need to change the tests for this rule (at the end of this file) to match the correct behaviour

bpandreotti (changes requested):

carcara/src/checker/rules/tautology.rs
Lachnitt: Sorry, I somehow missed your comment. Tests are changed.

PR #53, opened 2025-02-10
Drat and Drup checking

caotic123:
This PR adds DRAT and optionally DRUP checking rules. 

The rule is represented by:
`(step t4 (cl a) :rule drup :premises (t0 t1 t2) :args ((cl a b)))` or
`(step t4 (cl a) :rule drat :premises (t0 t1 t2) :args ((cl a b)))
`

Where args is the implied clause as usual.

### DIFF
--- a/carcara/src/drup.rs
+++ b/carcara/src/drup.rs
@@ -1,0 +1,329 @@
+use crate::ast::*;
+use indexmap::IndexSet;
+use std::borrow::{Borrow, BorrowMut};
+use std::collections::HashMap;
+use std::hash::{DefaultHasher, Hash, Hasher};
+use thiserror::Error;
+
+#[derive(Debug)]
+pub enum Implied<T, X> {
+    Pivot(T, X),
+    Bottom(X),
+    NotUnsat(),
+}
+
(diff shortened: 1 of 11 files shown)


carcara/src/drup.rs
```diff
@@ -0,0 +1,331 @@
+use crate::ast::*;
+use indexmap::IndexSet;
+use std::borrow::{Borrow, BorrowMut};
+use std::collections::HashMap;
+use std::hash::{DefaultHasher, Hash, Hasher};
```
bpandreotti, line 5: The CI build is failing because it uses rust version 1.72, which doesn't have `std::hash::DefaultHasher` ([apparently it was added in 1.76](https://blog.rust-lang.org/2024/02/08/Rust-1.76.0.html#stabilized-apis)). Instead, we should use `std::collections::hash_map::DefaultHasher`.

```suggestion
use std::collections::hash_map::DefaultHasher;
use std::collections::HashMap;
use std::hash::{Hash, Hasher};
```

bpandreotti (approved):

PR #57, opened 2025-03-27
Add BitVector Operation Support and Enhancements

bernborgess:
# Add BitVector Operation Support and Enhancements

This PR introduces new operators and macros to handle various BitVector (BV) operations, expands existing functionality, and improves test visibility.

### New BitVector Operators Added
- **Signed Comparisons**: `bvslt`, `bvsle`, `bvsgt`, `bvsge`
- **Unsigned Comparisons**: `bvugt`, `bvuge`, `bvule`
- **Bitwise Operations**: `bvxor`, `bvand`
- **Term Constructors**: `pbbterm` (pseudo-boolean bitvector constructor) and `int_of` (pseudo-boolean version of `bit_of`, for indexing)

### Key Enhancements
1. **Macro Expansions**:
   - Updated `match_term!` macro to support new BV operators.
   - Added `ParamOperator::BvIntOf` for BV-to-integer conversion.
   - Enhanced term construction with `build_term!` for new operators.

2. **Parser & Type Checking**:
   - Added handling for `BvPBbTerm` and `BvIntOf` in parser logic.

3. **Term Utilities**:
   - Added `Term::as_bitvector()` to extract BV constants.

4. **Testing & Diagnostics**:
   - Integrated `colored` crate for clearer test result output (✅/❌ with colors).
   - Enhanced error messages in test failures for easier debugging.
   - Added `colored = "2.1.0"` for colored terminal output in tests.
   - Updated contributor list.

5. **Linting and Code Style**:
   - Changes in lifetimes and other rust code as advised by `clippy`
   - Fixed some typos

## Impact
- Enables the creation of the `cutting-planes` and `pb-blasting` rules that use these operators (e.g., signed/unsigned comparisons, bitwise ops).
- Facilitates BV-to-integer conversions via `int_of`.
- Improves developer experience with better test feedback.


carcara/src/checker/rules/mod.rs
```diff
@@ -193,12 +195,42 @@ fn run_tests(test_name: &str, definitions: &str, cases: &[(&str, bool)]) {
...
+            Err(crate::Error::Checker { inner, rule, step }) => match inner {
+                CheckerError::Explanation(s) => s.clone(),
+                _ => format!("[{rule} at {step}] {:?}", inner),
+            },
+            Err(crate::Error::DoesNotReachEmptyClause) => {
+                "Proof does not reach empty clause".into()
+            }
+            Err(e @ crate::Error::Parser(_, _)) => format!("Parser error: {}", e),
+            Err(e) => format!("Unexpected error: {:?}", e),
+        };
```
bpandreotti, line 212: I'm curious, why the different behaviour for the `Explanation` error variant? Also, I think it would be simpler to just format the `crate::Error` directly, instead of deconstructing it:
```suggestion
        let error_message = match &check_result {
            Ok(_) => String::new(),
            Err(e) => format!("{}", e),
        };
```
(are you doing it this way to get a more concise error message?)

bernborgess: Ah, this "Explanation" error type is something Haniel showed me from his branch, that allows to express one of a kind error messages with an explanation. How could I do this in a function that may return a `RuleResult` ?

bpandreotti: Oh, I see. I think the existence of this error type makes sense then. Still, for the snippet I commented on, I think we can just format the error directly instead of matching on it.

bernborgess: Ok, how would you do it with the `Explanation` variant?
Do you prefer that I commit this change?

bpandreotti: Yes. This suggestion just simplifies the code a bit, it will still format the `Explanation` errors correctly. More precisely, when you define the `Explanation` variant you used `#[error("{0}")]`, so if you print the `crate::Error` directly, it will print the `CheckError`, which will automatically print the contents of the `Explanation` error.

bpandreotti (approved):

PR #63, opened 2025-04-07
Implement Cutting Planes Saturation Rule Checker (`cp_saturation`)

bernborgess:
# Implement Cutting Planes Saturation Rule Checker (`cp_saturation`)

This PR implements the `cp_saturation` rule checker for cutting planes proofs with comprehensive validation logic.

### Core Implementation
- Added `cp_saturation` checker that:
  - Reuses the helper functions
  - Ensures conclusion matches minimum between premise and the constant
 
- Change the addition tests and `unwrap_pseudoboolean_inequality` to account for a summation list of two elements or more differently than a single term (no summation)

PR #64, opened 2025-04-07
 Implement Cutting Planes Division Rule Checker (`cp_division`)

bernborgess:
# Implement Cutting Planes Division Rule Checker (`cp_division`)

This PR implements the `cp_division` rule checker for cutting planes proofs with comprehensive validation logic.

### Core Implementation
- Added `cp_division` checker that:
  - Reuses the helper functions
  - Ensures conclusion matches ceiling division of premise by the scalar

- Depends on #63

### DIFF
--- a/carcara/src/checker/rules/cutting_planes.rs
+++ b/carcara/src/checker/rules/cutting_planes.rs
@@ -217,8 +217,45 @@ pub fn cp_multiplication(RuleArgs { .. }: RuleArgs) -> RuleResult {
     Err(CheckerError::Unspecified)
 }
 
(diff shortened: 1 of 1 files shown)


carcara/src/checker/rules/cutting_planes.rs
bpandreotti, line 1: The rule is panicking if the provided argument is 0. You should explicitly check if it is, and if so return an error (something like `CheckerError::DivOrModByZero`). It would also be nice to add a test to cover this case.

bernborgess: Added this check

bpandreotti (approved):

PR #66, opened 2025-04-15
[rules] [strings] Add support for `re_unfold_pos` rule

vinisilvag:
This PR adds the `re_unfold_pos` rule from the Strings theory. More than that, this PR also splits the reasoning previously handled by a single rule into two distinct rules: `re_concat_unfold_pos` and `re_kleene_star_unfold_pos`. This was done to distinguish two behaviors that are notably different but were being implemented within the same rule. With this separation, even though the rules still share some logic and reasoning, it becomes clearer what each of them is doing specifically.

Rule reference can be found in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/api/cpp/enums/proofrule.html#_CPPv4N4cvc59ProofRule13RE_UNFOLD_POSE) and in the [CPC specification](https://github.com/cvc5/cvc5/blob/c56a15e30823c8360d7d7498d42db66d9c879f1d/proofs/eo/cpc/rules/Strings.eo#L299-L332). This rule also relies on a Skolem definition that has been rewritten to use only built-in operators (such as `Choice`, `Exists`, and other operators specific to the theory of Strings). The Skolem definition is available in the [cvc5 documentation](https://cvc5.github.io/docs-ci/docs-main/skolem-ids.html#_CPPv4N4cvc58SkolemId23RE_UNFOLD_POS_COMPONENTE), and the mapping used to implement it in Carcara will soon be available in the [Alethe specification document](https://verit.loria.fr/documentation/alethe-spec.pdf).

### DIFF
--- a/carcara/src/checker/rules/strings.rs
+++ b/carcara/src/checker/rules/strings.rs
@@ -359,28 +359,6 @@ fn singleton_elim(pool: &mut dyn TermPool, r_list: Vec<Rc<Term>>) -> Rc<Term> {
     }
 }
 
-/// A function that simplifies the application of the and operator between a term and the boolean
-/// constant true or another logical operator (and, or).
-///
-/// Implementation based on the behavior described in the [Ethos user manual.](https://github.com/cvc5/ethos/blob/main/user_manual.md#list-computation-examples)
-fn and_simplification(pool: &mut dyn TermPool, t: Rc<Term>, m: Rc<Term>) -> Rc<Term> {
-    if t.is_bool_true() {
-        return m;
-    }
-    if m.is_bool_true() {
-        return pool.add(Term::Op(Operator::And, vec![t]));
-    }
-    match m.as_ref() {
-        Term::Op(Operator::And, args) => {
-            let mut new_args: Vec<Rc<Term>> = Vec::new();
-            new_args.push(t);
-            new_args.extend(args.clone());
-            pool.add(Term::Op(Operator::And, new_args))
-        }
-        _ => t,
-    }
-}
-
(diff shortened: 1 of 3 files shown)


carcara/src/checker/rules/strings.rs
```diff
@@ -327,13 +348,175 @@ fn extract_arguments(t: &Rc<Term>) -> Result<Vec<Rc<Term>>, CheckerError> {
...
+    match m.as_ref() {
+        Term::Op(Operator::And, args) => {
+            let mut new_args: Vec<Rc<Term>> = Vec::new();
+            new_args.push(t);
+            new_args.extend(args.clone());
+            pool.add(Term::Op(Operator::And, new_args))
+        }
+        _ => t,
+    }
+}
```
bpandreotti, line 382: I'm a bit confused by the behaviour of this function.

Currently it is only used in one place (line 1259), and it is always called with `t` being an equality, and `m` being either the constant `true` or an `and` term. This means that the first `if` statement (`if t.is_bool_true()`) is never entered and can be removed; and the last `match` arm is also unreachable, and could be replaced with `_ => unreachable!()` for clarity. Is this because you plan to call this function with different values of `t` and `m` in the future?

Also about the last match arm, I don't understand why you would just return `t` if `m` is neither `true` nor an `and` term. In that case, I would expect the function to return something like `(and t m)`.

vinisilvag: This function is really a bit confusing. It's quite generic, but it was being used in a very specific context. I discussed this offline with @HanielB, and we agreed it would be better to move it into the code where it's actually being called and refactor it to better fit the specific use cases it was serving.

I'll commit these changes along with some documentation updates.

PR #70, opened 2025-04-23
Implement the rule `pbblast_bvule`

bernborgess:
This PR adds the implementation of the `pbblast_bvule` rule for checking pb bit blasted <= operation.
- Its similar to the `pbblast_bvult` rule but we use a different constant $0$ instead

PR #72, opened 2025-04-24
Implement the rule `pbblast_bvuge`

bernborgess:
This PR adds the implementation of the `pbblast_bvuge` rule for checking *unsigned* pb bit blasted >= operation.
- Its similar to the `pbblast_bvult` rule but we swapped the $x$ and $y$ sides and changed the constant to $0$
- Depends on #70


carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -176,8 +176,22 @@ pub fn pbblast_bvugt(RuleArgs { pool, conclusion, .. }: RuleArgs) -> RuleResult
...
+    // Get the summation lists
+    let sum_x = get_pbsum(sum_x);
+    let sum_y = get_pbsum(sum_y);
+
+    // Check that the constant is 0
+    let constant: Integer = constant.as_integer_err()?;
+    rassert!(
+        constant == 0,
+        CheckerError::Explanation(format!("Non-zero constant {}", constant))
+    );
```
bpandreotti, line 192: You don't need to check that the constant is 0 manually, `match_term_err!` can do that for you:
```suggestion
    let ((x, y), ((sum_x, sum_y), ())) =
        match_term_err!((= (bvuge x y) (>= (- sum_x sum_y) 0)) = &conclusion[0])?;

    // Get the summation lists
    let sum_x = get_pbsum(sum_x);
    let sum_y = get_pbsum(sum_y);
```

bpandreotti (approved):

PR #75, opened 2025-04-30
Adding parametric sorts

HanielB:
In preparation for supporting datatypes but will also be needed for supporting Rare rules.

### DIFF
--- a/carcara/src/ast/term.rs
+++ b/carcara/src/ast/term.rs
@@ -577,8 +577,10 @@ impl From<SortedVar> for Term {
 }
 
 impl Sort {
-    // Whether this sort can be unified with another. The map argument
-    // will be a substitution of sort variables to sorts
+    // Whether this sort can be matched with another, i.e., whether we
+    // can find a substitution to the sort variables of `self` that
+    // will make it equal to `target`. The map argument will store the
+    // substitution
     pub fn match_with(&self, target: &Sort, map: &mut IndexMap<String, Sort>) -> bool {
         match (self, target) {
             (Sort::Var(a), _) => {


HanielB:
I forgot there were some outstanding TODOs, will address them

carcara/src/ast/printer.rs
```diff
@@ -520,6 +520,11 @@ impl fmt::Display for Sort {
             Sort::Real => write!(f, "Real"),
             Sort::String => write!(f, "String"),
             Sort::RegLan => write!(f, "RegLan"),
+            Sort::Var(name) => write!(f, "{}", name),
+            Sort::ParamSort(args, s) => {
+                let par = format!("(par {:?} {})", args, s);
+                write!(f, "{}", par)
+            }
```
bpandreotti, line 527: You can pass these format arguments to `write!` directly, there's no need to first `format!` them into a string:
```suggestion
            Sort::ParamSort(args, s) => write!(f, "(par {:?} {})", args, s),
```

carcara/src/ast/term.rs
```diff
@@ -569,6 +576,68 @@ impl From<SortedVar> for Term {
...
+                match map.entry(a.to_string()) {
+                    Entry::Vacant(e) => {
+                        e.insert(target.clone());
+                    }
+                    Entry::Occupied(e) => {
+                        return e.get() == target;
+                    }
+                }
+                true
+            }
```
bpandreotti, line 594: I think you need to check if you are unifying a variable with itself, and if so, don't add anything to the map.

For instance, if you run this code to unify the sorts `(Array X Int)` and `(Array X X)` (where X is a sort variable), first it will add `X -> X` to the map, and then it will fail when trying to unify `X` and `Int`.

So maybe you should add a match case like this:
```rust
(Sort::Var(a), Sort::Var(b)) if a == b => true,
```

carcara/src/ast/term.rs
```diff
@@ -569,6 +576,68 @@ impl From<SortedVar> for Term {
...
+                    let matching = sorts_a
+                        .iter()
+                        .zip(sorts_b.iter())
+                        .filter(|&(t_a, t_b)| {
+                            let s_a = t_a.as_sort().unwrap();
+                            let s_b = t_b.as_sort().unwrap();
+                            s_a.match_with(s_b, map)
+                        })
+                        .count();
+                    matching == sorts_a.len() && matching == sorts_b.len()
```
bpandreotti, line 608: Here, you can use [`Iterator::all`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all):
```suggestion
                    sorts_a.iter().zip(sorts_b.iter()).all(|(t_a, t_b)| {
                        let s_a = t_a.as_sort().unwrap();
                        let s_b = t_b.as_sort().unwrap();
                        s_a.match_with(s_b, map)
                    })
```

carcara/src/ast/term.rs
```diff
@@ -569,6 +576,68 @@ impl From<SortedVar> for Term {
...
+            }
+            (Sort::Function(sorts_a), Sort::Function(sorts_b)) => {
+                for (a_t, b_t) in sorts_a.iter().zip(sorts_b.iter()) {
+                    let a_s = a_t.as_sort().unwrap();
+                    let b_s = b_t.as_sort().unwrap();
+                    if !a_s.match_with(b_s, map) {
+                        return false;
+                    }
+                }
+                true
```
bpandreotti, line 619: I think you can also use `Iterator::all` here.
```suggestion
                sorts_a.iter().zip(sorts_b.iter()).all(|(a_t, b_t)| {
                    let a_s = a_t.as_sort().unwrap();
                    let b_s = b_t.as_sort().unwrap();
                    a_s.match_with(b_s, map)
                })
```

bpandreotti (changes requested):

carcara/src/ast/term.rs
HanielB: Cool!

carcara/src/ast/term.rs
HanielB: You're right, but there is a complication... which is that this method is only *matching*, i.e., trying to find a substitution for `self` to match `target`. In that sense the example you gave should *not* match. 

For the purpose of parsing datatypes, which motivated these changes originally, matching suffices, since `target` is always ground. 

I think then that we could either:
- keep things as they are and better document that this is only doing matching
- generalize this method to actually unify the sorts, which would require some changes in it

What do you think?

bpandreotti: Ah, that makes sense. In that case, I think (at least for now) we don't need to bother implementing full unification. So I'd say I prefer going with the first option.

HanielB: Sounds good. I've update the doc

bpandreotti (approved):

PR #80, opened 2025-05-08
Implement rule `pbblast_bvand`

bernborgess:
Add the implementation of the Pseudo-Boolean blasting of `bvand`

- Extends the macro to accommodate `choice` (will also be used in `bvxor`)
- Support both normal and short-circuited bitvectors
- Unit tests with 1, 2 and 8 bit bitvectors, and short-circuited

> :warning:  Depends on checking if the binder is correct `z`

### DIFF
--- a/carcara/src/checker/rules/pb_blasting.rs
+++ b/carcara/src/checker/rules/pb_blasting.rs
@@ -1,7 +1,7 @@
 use super::{RuleArgs, RuleResult};
 use crate::{
     ast::{Rc, Sort, Term, TermPool},
-    checker::error::CheckerError,
+    checker::error::{CheckerError, EqualityError},
 };
 use rug::Integer;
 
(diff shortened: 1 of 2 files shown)


carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -375,14 +375,92 @@ pub fn pbblast_pbbconst(RuleArgs { conclusion, .. }: RuleArgs) -> RuleResult {
...
+        (0..n)
+            .map(|i| {
+                build_term!(
+                    pool,
+                    ((_ int_of { pool.add(Term::new_int(i)) }) { bv.clone() })
+                )
+            })
+            .collect()
+    }
+}
```
bpandreotti, line 399: This function is very similar to the existing function [`build_term_vec`](https://github.com/ufmg-smite/carcara/blob/a7888530241f85a9cfa849fa0f5e1b9047bb0fc2/carcara/src/checker/rules/bitvectors.rs#L8-L24). It would be nice if you could unify their implementations.

carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -375,14 +375,92 @@ pub fn pbblast_pbbconst(RuleArgs { conclusion, .. }: RuleArgs) -> RuleResult {
...
+        let (z_name, z_type) = &bindings[0];
+        assert!(z_name == "z");
+        assert!(*z_type.as_sort().unwrap() == Sort::Int);
+
+        // c1 : (>= @x0 z)
+        let (xic, zc) = match_term_err!((>= xi z) = c1)?;
+        rassert!(
+            xic == xi,
+            CheckerError::TermEquality(EqualityError::ExpectedEqual(xic.clone(), xi.clone()))
+        );
```
bpandreotti, line 430: Here, you can use the [`assert_eq` function](https://github.com/ufmg-smite/carcara/blob/a7888530241f85a9cfa849fa0f5e1b9047bb0fc2/carcara/src/checker/rules/mod.rs#L104-L113) from the `rules` module:
```suggestion
        super::assert_eq(xic, xi)?;
```
(or `assert_is_expected`, whichever you think is more appropriate -- the only difference is the error message)

carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -375,14 +375,92 @@ pub fn pbblast_pbbconst(RuleArgs { conclusion, .. }: RuleArgs) -> RuleResult {
...
+            zc.as_var() == Some(z_name) && pool.sort(zc) == *z_type,
+            CheckerError::Explanation(format!("Expected {z_name} but got {zc}"))
+        );
+
+        // c2 : (>= @y0 z)
+        let (yic, zc) = match_term_err!((>= yi z) = c2)?;
+        rassert!(
+            yic == yi,
+            CheckerError::TermEquality(EqualityError::ExpectedEqual(yic.clone(), yi.clone()))
+        );
```
bpandreotti, line 441: Same as previous comment.

carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -375,14 +375,92 @@ pub fn pbblast_pbbconst(RuleArgs { conclusion, .. }: RuleArgs) -> RuleResult {
...
+            CheckerError::Explanation(format!("Expected {z_name} but got {zc}"))
+        );
+        rassert!(
+            xic == xi,
+            CheckerError::TermEquality(EqualityError::ExpectedEqual(xic.clone(), xi.clone()))
+        );
+        rassert!(
+            yic == yi,
+            CheckerError::TermEquality(EqualityError::ExpectedEqual(yic.clone(), yi.clone()))
+        );
```
bpandreotti, line 460: Same as previous comment.

bpandreotti (changes requested):

carcara/src/checker/rules/pb_blasting.rs
bernborgess: The other function is in rules/bitvectors.rs. Should I keep it there and import directly in pb_blasting.rs or should we move it to another file that holds these helper functions? My concern is a change in that file impacting what I'm building on implicitly

bpandreotti: I think you can just import it directly. If something changes in that function, it will break your tests, so we will catch it.

bernborgess: Well, they are not the same due to the usage of `@int_of` instead of `@bit_of`. I think having the `@int_of` variant contained in the `pb_blasting.rs` will be better, since we may change this as well to finite fields later. @HanielB what do you think?

HanielB: I assume there is no way to parameterize the method with the operator and still use the macros, right, @bpandreotti ? If that's indeed the case then I think it's to have a very similar version of the same method in the PB module.

bpandreotti: Oh, my bad, I didn't catch the fact that the functions used different operators! Yes, in theory we could pass the operator as an argument, but then we would have to stop using the `match_term!` macro. But at that point, I don't think it's worth it to bother unifying them, so I think we leave it as it is.

PR #81, opened 2025-05-12
Devel/pb blasting rules pbblast bvxor

bernborgess:
Add the implementation of the Pseudo-Boolean blasting of `bvxor`

- Support both normal and short-circuited bitvectors
- Uses the `choice` macro
- Unit tests with 1, 2 and 8 bit bitvectors, and short-circuited

## Main Check:
- Looks for the pattern:

    $$(\texttt{bvxor}\ x\ y) \approx \texttt{pbbT} c_0 \dots c_{n-1}$$

    Each term $c_i$ is a **choice** application of form $\varepsilon z.\, \varphi[z]$
    applied to the conjunction of these constraints on the $i$ th bit:

    $$\textbf{x}_i+\textbf{y}_i \ge z $$
    $$z + \textbf{x}_i \ge \textbf{y}_i$$
    $$z + \textbf{y}_i \ge \textbf{x}_i$$
    $$2 \ge z + \textbf{x}_i + \textbf{y}_i$$

### DIFF
--- a/carcara/src/checker/rules/pb_blasting.rs
+++ b/carcara/src/checker/rules/pb_blasting.rs
@@ -255,8 +255,21 @@ pub fn pbblast_bvsgt(RuleArgs { pool, conclusion, .. }: RuleArgs) -> RuleResult 
 ///
 /// The expected shape is:
 ///    `(= (bvsge x y) (>= (+ (- x_sum (* 2^(n-1) x_n-1))) (- (* 2^(n-1) y_n-1) y_sum)) 0))`
-pub fn pbblast_bvsge(RuleArgs { .. }: RuleArgs) -> RuleResult {
-    Err(CheckerError::Unspecified)
+pub fn pbblast_bvsge(RuleArgs { pool, conclusion, .. }: RuleArgs) -> RuleResult {
+    let ((x, y), (((sum_x, sign_x), (sign_y, sum_y)), _)) = match_term_err!((= (bvsge x y) (>= (+ (- sum_x sign_x) (- sign_y sum_y)) 0)) = &conclusion[0])?;
+
+    // Get the summation lists
+    let sum_x = get_pbsum(sum_x);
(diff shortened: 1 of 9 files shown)


carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -399,8 +399,62 @@ fn get_bitvector_terms(bv: &Rc<Term>, pool: &mut dyn TermPool) -> Vec<Rc<Term>>
...
+    // Zip three lists into tuples
+    for ((bc, xi), yi) in bit_constraints.iter().zip(xs.iter()).zip(ys.iter()) {
+        let (bindings, (c1, c2, c3, c4)) = match_term_err!((choice ... (and c1 c2 c3 c4)) = bc)?;
+
+        // Single binding
+        assert!(bindings.len() == 1);
+        // Check z -> Int
+        let (z_name, z_type) = &bindings[0];
+        assert!(z_name == "z");
+        assert!(*z_type.as_sort().unwrap() == Sort::Int);
```
bpandreotti, line 418: One small problem: here, you are using `assert!`, instead of `rassert!`, which means this code will panic instead of returning a nice error message if the assertions are not valid.

For the first assertion, the parser already guarantees that there will be only one binding, so it's okay to use `assert!` (it's also okay to not check that at all, and just trust the parser). For the other two assertions, however, since they can be controlled by the user input, it would be better to use `rassert!`, and specify an error message.

This also affects the pull request for `pbblast_bvand` that was recently merged (my bad, I didn't catch that on the review), so please change that rule as well.

bpandreotti (changes requested):

carcara/src/checker/rules/pb_blasting.rs
bernborgess: Does it look better now?

bpandreotti (approved):

PR #86, opened 2025-05-20
Devel/pb blasting rules pbblast bvand ith bit

bernborgess:
Implementation of the rule `pbblast_bvand_ith_bit`, described [here](https://gitlab.uliege.be/verit/alethe/-/jobs/195555/artifacts/file/specification.pdf), page 59.

- Fixed minor typos
- Created unit tests
- Created helper `build_bvand_ith_bit_choice_term` that builds the `Term` used in place of `r`

Most notably, dared to extend the `build_term` macro, to accommodate the creation of a `choice` term. Feedback on how to separate the $(\texttt{and}\ c_1\ c_2\ c_3)$ from this case is very welcome!


carcara/src/ast/macros.rs
```diff
@@ -277,6 +277,14 @@ macro_rules! match_term_err {
...
     ($pool:expr, true) => { $pool.bool_true() };
     ($pool:expr, false) => { $pool.bool_false() };
+    // How to separate the (and c1 c2 c3) into another call to build_term
+    // so that an arbitrary term can sit there?
+    ($pool:expr, (choice (($z:literal $type:ty)) (and {$c1:expr} {$c2:expr} {$c3:expr}))) => {{
+        let int_sort = $pool.add($crate::ast::Term::Sort(Sort::Int));
+        let bindings = $crate::ast::BindingList(vec![($z.into(), int_sort.clone())]);
+        let body = build_term!($pool, (and {$c1} {$c2} {$c3}));
+        $pool.add(Term::Binder(Binder::Choice, bindings, body))
+    }};
```
bpandreotti, line 287: One way of doing it would be this:
```suggestion
    ($pool:expr, (choice (($z:literal $sort:ident)) $arg:tt)) => {{
        let sort = $pool.add($crate::ast::Term::Sort(Sort::$sort));
        let bindings = $crate::ast::BindingList(vec![($z.into(), sort)]);
        let body = build_term!($pool, $arg);
        $pool.add(Term::Binder(Binder::Choice, bindings, body))
    }};
```
Here we match a single arbitrary token tree as an argument (`$arg:tt`) and call `build_term!` recursively to build it.

Also, in the code you submitted, you are always using the `Sort::Int` sort instead of the sort provided to the macro. I changed it in this suggestion to take an identifier (`$sort:ident`), and construct the sort `Sort::$sort`. This will only work with the base variants of `Sort`, namely `Bool`, `Int`, `Real`, etc., but I think that's fine for now.

bernborgess: Thank you! I was trying to bind the types as well, I'll do the changes now

carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -574,3 +574,52 @@ pub fn pbblast_bvand(RuleArgs { pool, conclusion, .. }: RuleArgs) -> RuleResult
...
 }
+
+/// Helper function to create the `pbblast_bvand_ith_bit` choice term
+/// `(choice ((z Int)) (and (>= x z) (>= y z) (>= (+ z 1) (+ x y))))`
+fn build_bvand_ith_bit_choice_term(
+    pool: &mut dyn TermPool,
+    x: &Rc<Term>,
+    y: &Rc<Term>,
+) -> Rc<Term> {
+    let int_sort = pool.add(crate::ast::Term::Sort(Sort::Int));
```
bernborgess, line 585: @bpandreotti , Haniel and I were thinking how we could inject a variable into the `build_term!` to avoid such repetitive creations of `int_sort` and `z` in this case

carcara/tests/rules/pb_blasting.rs
```diff
@@ -3728,3 +3728,70 @@ fn pbblast_bvand_8() {
...
     }
 }
+
+#[test]
+fn pbblast_bvand_ith_bit() {
+    test_cases! {
+        definitions = "
+            (declare-const x Int)
+            (declare-const y Int)
+            ;; // ? Defining here doesn't let the definition be available for the rules...
```
bernborgess, line 3738: @bpandreotti, We also thought about the benefits of using `definitions` to store the `define-fun` term, and avoid repetition of it in each test case, right now all the test case would see if defined here is a `Var(r,Int)`

carcara/src/checker/rules/pb_blasting.rs
bpandreotti: I'm not exactly sure what you mean. Do you want to be able to pass a `SortedVar` to be used as the choice binding? E.g., something like
```rust
let var = ("z".into(), int_sort.clone());
build_term!(pool, (choice {var} {body}));
```

carcara/tests/rules/pb_blasting.rs
bpandreotti: This is a really good point. I've investigated this a bit, and the problem is that the rule tests use the default parser config:
https://github.com/ufmg-smite/carcara/blob/407d7aeb70462db75c01b4c7b3c5c67bb0bdabd2/carcara/tests/rules/mod.rs#L14
in which `apply_function_defs` is false. So if you replace this config with something like
```rust
parser::Config {
    apply_function_defs: true,
    ..Default::default()
}
```
it should work.

carcara/src/checker/rules/pb_blasting.rs
bernborgess: Not really, just some syntax in the macro to introduce a variable we could do something like
```rust
let term = build_term!(pool,(+ (let x Int) (let y Int)));
```
What do you think?

bpandreotti: Oh, I see. Yes, I think it makes sense to have support for this. One way of doing it would be to add a rule like this to the macro:
```rust
($pool:expr, (let $name:ident $sort:ident)) => {{
    let sort = $pool.add($crate::ast::Term::Sort(Sort::$sort));
    $pool.add($crate::ast::Term::new_var(stringify!($name), sort))
}};
```
(here we use `stringify!` so the identifier `x` becomes the string literal `"x"`)

carcara/tests/rules/pb_blasting.rs
bernborgess: Thanks! Added this setting in the last commit.

carcara/src/checker/rules/pb_blasting.rs
```diff
@@ -574,3 +574,39 @@ pub fn pbblast_bvand(RuleArgs { pool, conclusion, .. }: RuleArgs) -> RuleResult
...
 }
+
+/// This rule extracts assertions of the ith bit of an application of
+/// `pbblast_bvand`, given its arguments were x and y, we conclude
+///  `(>= x r) (>= y r) (>= (+ r 1) (+ x y)))`
+/// In which ri is the choice element from the pseudo boolean bit blasting
+/// of the bvand rule
+pub fn pbblast_bvand_ith_bit(RuleArgs { args, pool, conclusion, .. }: RuleArgs) -> RuleResult {
+    let x = &args[0];
+    let y = &args[1];
```
bpandreotti, line 585: You should make sure that exactly two arguments are passed, using `super::assert_num_args(args, 2)?;`. Currently, if no argument is passed to the rule, it will panic instead of returning an error.

bpandreotti (changes requested):

carcara/src/checker/rules/pb_blasting.rs
bernborgess: Good catch! I implemented the fix and some unit tests to catch it!

bpandreotti (approved):

PR #87, opened 2025-05-24
Add parse for Rare Rules

caotic123:


carcara/src/ast/term.rs
```diff
@@ -88,6 +88,7 @@ pub enum Sort {
     /// The associated term is the BV width of this sort.
     BitVec(Integer),
 
+    // TODO delete this and incorporate it to function sort?
```
HanielB, line 91:
```suggestion
```

carcara/src/lib.rs
```diff
@@ -148,9 +151,10 @@ pub fn check_parallel<T: io::BufRead>(
     use std::sync::Arc;
     let mut run_measures: RunMeasurement = RunMeasurement::default();
 
-    // Parsing
+    // Parsing (TODO : Complete rare rules)
```
HanielB, line 154: ?

HanielB (changes requested):
Some initial minor comments, as I'll leave @bpandreotti take a closer look. But you should already provide a proper description to the PR, specially characterizing what it is that you are parsing, what concrete syntax is expected, and what is the AST that the parser generates.

I also think that `rules` is not a good name for the AST module, without a qualifier, like "Rare" or "Rewrites", so the module would be "RareRules" or ¨RewriteRules".

carcara/src/checker/rules/mod.rs
```diff
@@ -155,6 +156,70 @@ fn assert_is_bool_constant(got: &Rc<Term>, expected: bool) -> RuleResult {
...
+        let definitions: &str = $defs;
+        $({
+            let name: &str = $name;
+            let cases = [ $(($proof, $exp),)* ];
+            $crate::checker::rules::run_tests(name, definitions, &cases);
+        })*
+    }};
+}
+
+// Since the rule submodules use the `test_cases` macro, we have to declare them here, after the
```
bpandreotti, line 222: I think you accidentally reverted some changes I made when doing a merge. This part should be removed, since it is now in `carcara/tests/rules/mod.rs`.
```suggestion
```

carcara/src/parser/error.rs
```diff
@@ -128,6 +128,26 @@ pub enum ParserError {
     /// The parser encountered an unknown qualified operator.
     #[error("not a valid qualified operator: '{0}'")]
     InvalidQualifiedOp(String),
+
+    /// The parser encountered an unknown qualified operator.
```
bpandreotti, line 132: This comment is wrong.

bpandreotti:
There are also a bunch of linter errors (which you can see with `cargo clippy`) and some formatting inconsistencies (which you can fix with `cargo fmt`).

carcara/src/parser/mod.rs
```diff
@@ -65,24 +66,38 @@ impl Config {
...
     parser.reset(proof)?;
     let proof = parser.parse_proof()?;
-    Ok((problem, proof))
+    if let Some(rules) = rules {
+        parser.reset(rules)?;
+        let rules = parser.parse_rare();
+        let rules = match rules {
+            Ok(t) => Ok(t),
+            Err(v) => {
+                println!("\x1b[1;31m[UNEXPECTED]\x1b[0m At \x1b[34m.rare\x1b[0m file");
```
bpandreotti, line 94: There's no need to print the error here. Since you return the error, it will eventually be printed in the main function.

carcara/src/parser/mod.rs
```diff
@@ -1689,6 +1708,47 @@ impl<'a, R: BufRead> Parser<'a, R> {
...
+                        (self.pool.add(var), value)
+                    })
+                    .collect();
+
+                let result = Substitution::new(self.pool, substitution)
+                    .unwrap()
+                    .apply(self.pool, &inner);
+
+                Ok(result)
+            }
```
bpandreotti, line 1751: I'm not sure what is the purpose of this code, it doesn't seem to be related to rare? Also, it seems to be duplicated further down (maybe this was an error when rebasing).

carcara/src/parser/rare/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+                },
+                false,
+            )?;
+            Ok(Body::Premise(terms))
+        }
+        attribute => Err(Error::Parser(
+            ParserError::InvalidRareFunctionAttribute(attribute.iter().collect()),
+            parser.current_position,
+        )),
+    }
```
bpandreotti, line 93: I think this can be simplified to:
```suggestion
    let qualified_arg = parser.expect_keyword()?;
    match qualified_arg.as_str() {
        "conclusion" => {
            let rewrite_term = parser.parse_term()?;
            Ok(Body::Conclusion(rewrite_term))
        }
        "args" => {
            fn parse_args<R: BufRead>(parser: &mut Parser<R>) -> CarcaraResult<Vec<String>> {
                parser.expect_token(Token::OpenParen)?;
                parser.parse_sequence(super::Parser::expect_symbol, false)
            }
            let args = parse_args(parser)?;
            Ok(Body::Args(args))
        }
        "premises" => {
            parser.expect_token(Token::OpenParen)?;
            let terms = parser.parse_sequence(
                |parser| {
                    let term = parser.parse_term()?;
                    Ok(term)
                },
                false,
            )?;
            Ok(Body::Premise(terms))
        }
        _ => Err(Error::Parser(
            ParserError::InvalidRareFunctionAttribute(qualified_arg),
            parser.current_position,
        )),
    }
```

carcara/src/parser/rare/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    let body = body.iter().fold(body_definitions, |mut body, x| {
+        match x {
+            Body::Conclusion(term) => body.conclusion = Some((*term).clone()),
+            Body::Premise(term) => body.premises = term,
+            Body::Args(args) => body.args = args,
+        }
+        body
+    });
+
+    if Option::is_none(&body.conclusion) {
```
bpandreotti, line 125: I think it's more idiomatic to use `is_none` like this:
```suggestion
    if body.conclusion.is_none() {
```

carcara/src/parser/rare/mod.rs
bpandreotti, line 1: In general, I think I prefer if the functions in this module were directly methods of the `Parser` type, instead of taking a `&mut Parser` as an argument. You can add new methods to parser in this module by having a separate `impl` block here:
```rust
impl<'a, R: BufRead> Parser<'a, R> {
    fn parse_parameters(&mut self) -> CarcaraResult<(String, TypeParameter)> {
        // ...
    }
    // ...
}
```

Yes, you can have multiple `impl` blocks for the same type, even on different modules!

bpandreotti (changes requested):
Aside from these issues, there are also some merge conflicts that should be fixed.

bpandreotti (approved):

PR #88, opened 2025-05-27
Implement rule `cp_literal`

bernborgess:
Implementation of the rule `cp_literal`, described [here](https://gitlab.uliege.be/verit/alethe/-/jobs/196565/artifacts/browse), page 54. "A literal 𝑙 being greater or equal to zero can be introduced at any time"

This rule is true when we suppose $l$ is either a pseudo boolean variable or the negation of one `(- 1 l)`

### DIFF
--- a/carcara/src/checker/mod.rs
+++ b/carcara/src/checker/mod.rs
@@ -495,6 +495,7 @@ impl<'c> ProofChecker<'c> {
             "cp_multiplication" => cutting_planes::cp_multiplication,
             "cp_division" => cutting_planes::cp_division,
             "cp_saturation" => cutting_planes::cp_saturation,
+            "cp_literal" => cutting_planes::cp_literal,
 
             "string_decompose" => strings::string_decompose,
             "string_length_pos" => strings::string_length_pos,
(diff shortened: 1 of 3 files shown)


bpandreotti (commented):
The implementation looks great, but I was wondering: is this rule even correct? Why can you simply conclude that $$l > 0$$ for any arbitrary integer $$l$$?

In fact, the way it's currently specified, you can use this rule to conclude that $$-1 >= 0$$:
```(step t1 (cl (>= (- 1) 0)) :rule cp_literal :args ((- 1)))```

bernborgess:
> The implementation looks great, but I was wondering: is this rule even correct? Why can you simply conclude that l > 0 for any arbitrary integer l ?
> 
> In fact, the way it's currently specified, you can use this rule to conclude that − 1 >= 0 : `(step t1 (cl (>= (- 1) 0)) :rule cp_literal :args ((- 1)))`

@HanielB how can I put it, in the future we going to change the type of the parameter to be finite field 2 which means the value can only be zero or one but for now we're using integers as a replacement.

bpandreotti:
Ah, that makes sense. Thanks for the clarifications!

PR #97, opened 2025-06-30
Add buffer_entire_file option

netolcc06:


carcara/src/parser/mod.rs
```diff
@@ -49,6 +49,11 @@ pub struct Config {
     /// If `true`, the parser will parse arguments to the `hole` rule, expecting them to be valid
     /// terms.
     pub parse_hole_args: bool,
+
+    /// If `true`, the parser will buffer the entire file content in memory before parsing
+    /// instead of reading line-by-line. This can improve performance in network file systems
+    /// or cluster environments at the cost of increased memory usage.
+    pub buffer_entire_file: bool,
```
bpandreotti, line 56: Since you are handling this logic in the `get_instance` function in the `cli` crate (which I think is better, to be honest), you don't need to have this option in `ParserOptions`, as it's not used. I think we should remove it.

bpandreotti (approved):

PR #98, opened 2025-06-30
Extend slice to work in subproofs and produce problem files

jjakpor:
This pull request contains some extensions to slice.
- Most importantly, slice can now slice steps within subproofs.
- Slice now produces problem files. The output problem file contains the logic string and sort and function declarations of the original problem, an assertion corresponding to each assumption in the sliced problem, and `(assert false)` to be resolved trivially with `(cl (not false))`.
- Slice now writes its output to files specified by the user or to default paths in the working directory.


jjakpor:
I converted this back to a draft so that I can add the tests and make a few more changes that were suggested recently.

PR #99, opened 2025-06-30
Emit a parser error for assume commands after step commands within subproofs

jjakpor:
The Alethe specification says on page 11 that a proof is not allowed to issue assume commands after the first step command of a subproof has been issued. However, there was not a parser error for this. I added a new parser error type to the parser error enum and changed the subproof stack in the parser to keep track of whether the first step has been issued at every level, not just the top. This does cause `ast::tests::test_node` to fail though because the proof node representation permutes the subproof premises.


jjakpor:
By the way, I'm not really saying we should merge these changes right now, especially since their interaction with the proof node permutation might take some time to address. Before I saw that complication, I thought it would be a quick fix, which is why I made the edits.

jjakpor:
Thanks for your response! I'll switch it from a parser error to a warning when I get a chance.

jjakpor:
I changed it to a warning instead of an error now.

bpandreotti (approved):

PR #101, opened 2025-07-22
Emit parser error when assumption appears after step in subproof

jjakpor:
Now that the order of premises in subproofs [has been fixed](https://github.com/ufmg-smite/carcara/commit/93eb43a558c8311999d065516138bbc7c34482aa), I thought we could emit a parser error now as in the original version of PR #99 without testing errors (unless there's another reason to keep it as a warning).


bpandreotti:
Thanks for the contribution!

PR #104, opened 2025-11-19
Add Lambdapi translation module

NotBad4U:
Dear all,

This PR adds support in Carcara for translating Alethe proofs into Lambdapi. The new translation module takes an Alethe proof AST produced by cvc5, translates it into a Lambdapi AST, and then emits a corresponding Lambdapi script that can be type-checked independently. This is intended to enable proof reconstruction experiments in Lambdapi, facilitate cross-checking SMT proofs with a small trusted kernel, and serve as a basis for further integrations with proof assistants.

To be fully integrated, we should address the following points:

- [ ] Add end-to-end tests for the Alethe → Lambdapi translation @bpandreotti
    - Maybe read env var in `test_example_files.rs` to get lambdapi binary
- [ ] Decide whether to include Lambdapi + Lambdapi- stdlib as a submodule/dependency so users can not only translate but also check proofs
  - Add a shell script
- [ ] Define and document the CLI / API interface for translation @bpandreotti 
      (flags, output path, format, etc.)
- [ ] Define and document the supported Alethe / Lambdapi versions and what happens with unsupported features. Maybe we need to fix a version of cvc5 for the translation.
- [ ] Improve error reporting and diagnostics for translation failures
- [ ] (low priority) Add regression tests to avoid translation breaks because of a new feature added.
- [ ] Refactor and clean up the implementation
    - Move lambdapi folder at the root, and document the `make install` 
- [ ] Add examples in the repository (small Alethe proofs with their Lambdapi output)
- [ ] Update the documentation to describe the new feature and how to use it @bpandreotti 
- [ ] Integrate the feature into CI (run translation + optional check on a small suite) @bpandreotti 
- [x] Fix all compilation warnings
- [ ] Fix clippy errors

Feedback on the overall design of the translation backend, naming, and integration points is very welcome. In particular, I would appreciate comments on whether the current structure feels sufficiently modular and on any improvements you see for the interface or error reporting. I am happy to revise the implementation to better align with Carcara’s style and long-term roadmap.


HanielB:
> Awesome! I would be interested in reviewing the code, as I am also implementing 2 others compilers for Alethe, and it would be nice to share some infrastructure with yours. Since there are some TODOs left before your work can be integrated, I am wondering if we could also consider as another TODO converging to some common "translation" infrastructure. In my case, I have been working on the [translation branch](https://github.com/ufmg-smite/carcara/tree/translation), where I tried to share stuff between a compiler into Eunoia and another into TSTP (in the `translation` module). If you agree, as a first step I could look into your `lambdapi` module in general, to have a taste of the general architecture of the compiler, looking for opportunities for sharing structure with my compilers.

I discussed a bit with @bpandreotti this point and our view is that in the interest of time (so that we could hopefully finish this PR by the end of @NotBad4U's visit) it'd be best to initially get this work into main (after the necessary modifications and so on) and then once that is done work towards generalizing things and synchronizing so that what you've been working on can be combined well with this, @Mallku2. 

Of course as you say you can already start getting acquainted with this work to facilitate the future convergence, as you say. @NotBad4U is gonna be visiting UFMG until end of next week.

Issue #118, opened 2026-04-07
Improve ergonomics of `match_term!` macros


bpandreotti:
Currently, the return type of the `match_term!` and `match_term_err!` macros is as nested tuple that reflects the structure of the pattern given. For example, for the call
```rust
match_term!((and (= x y) (or a b c)) = term)
```
the return type will be `((&Rc<Term>, &Rc<Term>), (&Rc<Term>, &Rc<Term>, &Rc<Term>))`. Usually the user of the macro immediately destructures this tuple in a let binding, like:
```rust
let ((x, y), (a, b, c)) = match_term!((and (= x y) (or a b c)) = term)
```

The problem is that, for more complicated patterns, finding out the correct destructuring to get the matched terms can be an annoying and sometimes error-prone process. This could be improved if instead the macro always returned a flat tuple (like `(&Rc<Term>, &Rc<Term>, &Rc<Term>, &Rc<Term>, &Rc<Term>)`) or a constant-sized array (like `[&Rc<Term>; 5]`).

Due to how the code is generated in the macro (and the fact that macros can't inspect type information), it seems to me that the best approach would be to have the macro use a separate function that does that flattening, but that in itself [seems difficult](https://stackoverflow.com/questions/40178572/are-there-any-ways-to-recursively-flatten-tuples).

A different but realted improvement to the macro ergonomics would be to check that terms are equal when they are given the same name. For example, the call `match_term!((= x x) = term)` would check that the two sides of the equality are the same terms, and return `None` if not.

PR #122, opened 2026-04-13
Flattening macros

netolcc06:


carcara-macros/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+        TokenTree::Literal(lit) => {
+            let s = lit.to_string();
+            if s == "\"\"" {
+                Pat::EmptyString
+            } else {
+                let n: u64 = s.parse().unwrap_or_else(|_| {
+                    panic!("unsupported literal in match_term_flat!: {s}")
+                });
+                Pat::Literal(n)
+            }
```
bpandreotti, line 156: While we're at it, it might be nice to add support for other types of literals (e.g. real literals like 1.0)

carcara-macros/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+                    assert!(
+                        tokens.len() >= 5,
+                        "binder pattern must be ({name} ... <body>)"
+                    );
+                    let inner_tt = tokens[tokens.len() - 1].clone();
+                    return Pat::Binder {
+                        binder: name,
+                        inner: Box::new(parse_pat(inner_tt, ctr)),
+                    };
+                }
```
bpandreotti, line 190: I think this isn't actually checking that what follows the binder is `...`. So something like `(forall blah t)` would be accepted. It also doesn't ensure the length is exactly 5, so extra tokens (`(forall foo bar t)`) also work.

carcara-macros/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+                        let op_args: Vec<Pat> = op_arg_tokens
+                            .iter()
+                            .cloned()
+                            .map(|t| parse_pat(t, ctr))
+                            .collect();
+                        let args: Vec<Pat> = tokens[1..]
+                            .iter()
+                            .cloned()
+                            .map(|t| parse_pat(t, ctr))
+                            .collect();
```
bpandreotti: It seems like this doesn't support the variadic arguments pattern (`...`) with ParamOps, e.g. `((_ extract 3 2) ...)`. This is a bit of an edge case, but it still would be nice to support it. You should move the argument parsing logic below to a separate function, and reuse it for ParamOp and regular operators.

bpandreotti (commented):
Overall the code looks good, save for a few minor things.

One thing that is missing is that this PR does not actually change all uses of `match_term` to use the new implementation. On that note, instead of naming the new implementation `match_term_flat_err`, I think we should keep the name `match_term(_err)!`, and just change all uses to use the flattened version.

Note that this PR also doesn't add support for checking that terms are equal when given the same name, as mentioned in #118. I think it's best to add that in a separate PR, after we merge this one.

Another minor thing is that the code is failing CI because of linting and formatting errors, which should be easy to fix.

carcara/src/checker/rules/polynomial.rs
bpandreotti: This rule was removed in 2faf02ca5a7a1e1af1930a133ecf4b56b1523542. You probably added it back by accident when merging.

carcara/src/checker/shared.rs
```diff
@@ -202,8 +202,8 @@ pub fn get_rule_shared(rule_name: &str, elaborated: bool) -> Option<crate::check
         "la_disequality" => linear_arithmetic::la_disequality,
         "la_totality" => linear_arithmetic::la_totality,
         "la_tautology" => linear_arithmetic::la_tautology,
-        "poly_simp" => polynomial::poly_simp,
-        "bv_poly_simp" => polynomial::bv_poly_simp,
+        // We allow the usage of legacy rules `bv_poly_simp(_eq)`
+        "poly_simp" | "bv_poly_simp" => polynomial::poly_simp,
         "poly_simp_rel" => polynomial::poly_simp_rel,
         "bv_poly_simp_eq" => polynomial::bv_poly_simp_eq,
```
bpandreotti: See previous comment.

bpandreotti (changes requested):

bpandreotti (approved):

PR #123, opened 2026-06-09
Fix rare-list checking: parameterize RareList sort by element type

caotic123:
Split off from #111 — first of two fixes for rare-list checking.

## What
Change `Sort::RareList` from a unit variant to `Sort::RareList(Rc<Term>)` so a rare list carries the sort of its elements. Sort computation, sort comparison (`polyeq` / `match_with`), substitution, and the parser now look through a rare list to its element sort. This is needed for:

- bitvector width inference (concat, extract, zero/sign-extend, repeat, …),
- array `select` / `store` result sorts,
- polymorphism checks,
- `:list` parameters in RARE rules.

It also splices many-term reconstruction results into operator arguments in `reconstruct_meta_terms`, which is required to expand `:list` arguments when reconstructing a rewritten term.


carcara/src/ast/pool/mod.rs
```diff
@@ -137,7 +137,11 @@ impl PrimitivePool {
                 | Operator::BvSRem
                 | Operator::BvSMod
                 | Operator::BvAShr => {
-                    match self.compute_sort(&args[0]).as_sort().unwrap().clone() {
+                    let sort = match self.compute_sort(&args[0]).as_sort().unwrap().clone() {
+                        Sort::RareList(inner) => inner.as_sort().unwrap().clone(),
+                        sort => sort,
+                    };
```
bpandreotti: Since you use this logic in many places, it might be good to put it into a method in `Sort`. Specifically, a method that unwraps the sort if it is `Sort::RareList`, or just returns it otherwise.

bpandreotti (changes requested):
There are still merge conflicts that need to be resolved, and one other minor change I commented, but overall it looks good.

bpandreotti (approved):

PR #124, opened 2026-06-09
Normalize rare lists for unary operations in rewrite engine

caotic123:
Split off from #111 — second of two fixes for rare-list checking.

## What
Adds the missing `get_rules` equations needed to normalize a rare list that appears as the sole argument of a unary `and`/`or`, and to collapse a singleton `and`/`or` to its argument:

```
(And (RareList ..x..)) ~> (And x)
(Or  (RareList ..x..)) ~> (Or  x)
(And x) ~> x
(Or  x) ~> x
```

Reworks `rewrite_meta_terms` to drive these rules to a fixpoint: it splices many-term (rare list) results into the parent operator's arguments, re-normalizes the rewritten arguments, and memoizes results via a `RewriteContext` (with in-progress tracking to avoid re-entrancy), replacing the previous non-caching recursion.


bpandreotti (approved):
Looks great! But it's failing CI due to a formatting inconsistency, so please run `cargo fmt` to fix it.

bpandreotti (approved):

PR #125, opened 2026-06-09
Normalize all n-ary operators in the RARE rewrite engine

caotic123:
Broadens RARE meta-term normalization from `and`/`or` to **every n-ary (right-assoc-nil) operator** carcara represents.

> **Stacked on #123** (rare-list checking / parameterized `Sort::RareList`). The bitvector cases cannot even be *built* without #123's `compute_sort` unwrapping, so this builds on it and also carries #124's And/Or engine refactor (which it supersedes). Until #123 lands, the diff here will also show #123's changes.

## What

`get_rules()` now carries, per operator and **in this order**, a flatten rule `(Op (RareList ..x..)) ~> (Op x)`, a singleton rule `(Op x) ~> x`, and — where the nil terminator is a static term — an empty rule `(Op) ~> <nil>`:

- Arithmetic: `+` (nil `0`), `*` (nil `1`)
- Strings: `str.++` (nil `""`)
- Regex: `re.++`, `re.union` (nil `re.none`), `re.inter` (nil `re.all`)
- Bitvector: `bvand`, `bvor`, `bvxor`, `bvadd`, `bvmul`, `concat`

Flatten must precede singleton — `check_rewrites` returns the first match and `(Op x)` also matches a single rare-list argument, so flatten-first is what splices a list instead of collapsing it.

### Width-dependent bitvector nils

Bitvector nil terminators depend on the operand width and can't be static constants. They're synthesized in `bv_nil_if_empty`: when an n-ary bitvector op's argument list becomes empty, recover the width and build the nil (`0` for bvadd/bvor/bvxor, all-ones for bvand, `1` for bvmul, the zero-width empty bitvector for concat). When the width is genuinely unrecoverable (a bare operator with no operands), the term is left un-normalized rather than guessing.

## Known limitations

- `re.++` empty case omitted (its nil `(str.to_re "")` is not a zero-arg term; flatten + singleton suffice).
- `(+)`/`(*)` empties emit integer `0`/`1`; Real-sorted empties would want `0.0`/`1.0` (only matters in the all-empty degenerate case).
- `set.*` / `Tuple` are out of scope (absent from carcara's `Operator` enum).


carcara/src/rare/mod.rs
```diff
@@ -1,17 +1,73 @@
...
     vec![
         build_equation!((RareList ..x..) ~> x),
+        // Booleans
+        build_equation!((And (RareList ..x..)) ~> (And x)),
+        build_equation!((Or (RareList ..x..)) ~> (Or x)),
+        build_equation!((And x) ~> x),
+        build_equation!((Or x) ~> x),
         build_equation!((Or true) ~> true),
-        build_equation!((Or) ~> true),
-        build_equation!((And) ~> false),
```
bpandreotti, line 24: Why were these two nil cases removed? From the comment above, I assumed they were required.

bpandreotti (changes requested):
Again, a few merge conflicts and formatting stuff. Otherwise, looks good to me.

PR #128, opened 2026-07-22
First PR for the compilers into Eunoia and TSTP

Mallku2:
This PR focuses on the shared infrastructure among the compilation routines that I have implemented. There are some things that we might remove (explained below), but I think that I might benefit from having, first, some general review of the code. It contains:

- Definition of a simple `Translator` trait.
- Routines for representing "proof as lists of steps" and converting for converting a DAG representation into such list representation (PreOrderedAletheProof: this is something that I want to be sure that can be deleted).
- A notion of subproof scopes in Alethe, useful for the translation routines.
- The actual main infrastructure of a compiling routine (VecToVecTranslator), that can receive a DAG representation of a proof, convert it into a sequence of alethe steps, and then compile it into some specified target language.


carcara/src/translation/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+
+        // TODO: let Some(true)?
+        let context_introduced = self.context_introduced.pop();
+
+        if context_introduced == Some(true) {
+            // We are closing a context (instead of closing the scope of some other
+            // binder).
+            self.contexts_opened -= 1;
+        }
+    }
```
bpandreotti, line 115: It might be better to make this function explicitly panic in the case that `self.context_introduced` is empty, instead of silently failing. This can either be via an explicit `assert!(!self.context_introduced.is_empty())`, or by calling `unwrap` on the result of `self.context_introduced.pop()`.

Mallku2: > assert!(!self.context_introduced.is_empty())

nice catch!, forgot to be defensive over there. Done!

bpandreotti (approved):

PR #129, opened 2026-07-28
Add a new rule `eq_mp` and its elaboration

HanielB:
This commit adds a new extra rule, `eq_mp`, that represents a common pattern in Alethe proofs where a formula is used to derive another that is equivalent to it from a proof of this equivalence. An elaboration to the regular way to do this with current rules is also added. 

Specifically, the rule derives `F2` from `F1` and `(= F1 F2)` and the local elaboration pass replaces it with a resolution taking the original premises and a new `equiv_pos2` step. 

A guard is added for when `F2` is the negation of `F1`, in which case the two literals introduced by `equiv_pos2` collapse into one and the resolution would derive the empty clause, so the step is left as is.

Claude was used to write the code from my description of the rule and the elaboration, and I inspected the result, manually making the elaboration code a bit more succinct. 

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/carcara/src/elaborator/local/eq_mp.rs
+++ b/carcara/src/elaborator/local/eq_mp.rs
@@ -1,0 +1,59 @@
+use crate::{
+    ast::*,
+    elaborator::{error::ElaborationError, IdHelper},
+};
+
+/// Elaborates an `eq_mp` step into a `resolution` step taking the
+/// original premises and a new `equiv_pos2` step.
+///
+/// That is, a step of the form
+/// ```text
+/// (step t3 (cl F2) :rule eq_mp :premises (t1 t2))
+/// ```
+/// where `t1` concludes `F1` and `t2` concludes `(= F1 F2)`, becomes
+/// ```text
+/// (step t3.t1 (cl (not (= F1 F2)) (not F1) F2) :rule equiv_pos2)
+/// (step t3 (cl F2) :rule resolution :premises (t3.t1 t2 t1) :args ((= F1 F2) false F1 false))
+/// ```
+pub fn eq_mp(
+    pool: &mut PrimitivePool,
+    _: &mut ContextStack,
+    step: &StepNode,
+) -> Result<Rc<ProofNode>, ElaborationError> {
+    assert_eq!(step.clause.len(), 1);
+    assert_eq!(step.premises.len(), 2);
+    let (phi_1_step, equiv_step) = (&step.premises[0], &step.premises[1]);
+    assert_eq!(equiv_step.clause().len(), 1);
+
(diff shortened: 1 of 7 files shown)


carcara/src/elaborator/local/eq_mp.rs
```diff
@@ -0,0 +1,59 @@
...
+    let (phi_1, phi_2) = match_term_err!((= phi_1 phi_2) = &equivalence)?;
+    let (phi_1, phi_2) = (phi_1.clone(), phi_2.clone());
+
+    // If `phi_2` is exactly the negation of `phi_1`, the last two literals introduced by
+    // `equiv_pos2` are the same, and the resolution below would derive the empty clause instead of
+    // the step's conclusion. That can only happen if the premises are contradictory, in which case
+    // we simply leave the step as it is.
+    if match_term!((not phi) = &phi_2) == Some(&phi_1) {
+        return Ok(Rc::new(ProofNode::Step(step.clone())));
+    }
```
bpandreotti: If we really want to remove every instance of `eq_mp`, I think there's a way to elaborate this case instead of just giving up. Consider this example of this edge case:
```
(step h1 (cl p) :rule hole)
(step h2 (cl (= p (not p))) :rule hole)
(step t3 (cl (not p)) :rule eq_mp :premises (h1 h2))
```

We can derive the same `equiv_pos2` step as we do in the general case:
```
(step t3.t1 (cl (not (= p (not p))) (not p) (not p)) :rule equiv_pos2)
```
resolve it with only the second premise (`h2` in our case):
```
(step t3.t2 (cl (not p) (not p)) :rule resolution :premises (t3.t1 h2) :args ((= p (not p)) false))
```
and then apply a contraction:
```
(step t3 (cl (not p)) :rule contraction :premises (t3.t2))
```
And we get back the original conclusion, using only existing rules.

HanielB: I had not thought deeply about but I'd assumed there'd be a way to do this. Thanks for thinking of one! I had not investigated it further though because a proof that has a step concluding `(= p (not p))` is so silly that not elaborating this step seems like the least problem hehe. So I thought that having the simpler code would be preferable...  but it can be kept basically as simple as it is right now, since just dropping one of the premises is enough to have a valid Alethe proof, and a contraction step may be added later by the uncrowd pass (that required being careful with IDs though, because there is some dangerous interactions between the passes from what I could gather). 

PTAL

bpandreotti: That's a good point, I'm all in favor of ignoring these silly cases sometimes. Glad it could be kept simple, though!

bpandreotti (approved):

PR #131, opened 2026-07-30
alethe2eunoia and alethe2tstp: ASTs and pretty printers

Mallku2:
This PR adds basic mod infraestructure for the compilers alethe2eunoia and alethe2tstp:
- ASTs of Eunoia and TSTP.
- Pretty printers.
- Some changes in `translation/mod.rs`, required to better deal with ProofCommands.

Both modules added have a similar structure. The main difference resides in that alethe2eunoia requires some extra module, `alethe_signature`, that defines a mapping between the names of rules in Alethe, and their proper representation in Eunoia. This handles some minor differences between concrete syntax (e.g., the `cl` Alethe operator is denoted as `@cl` in our mechanization).

With regard to the pretty printers: I could totally reuse some other pretty printing infraestructure already present in Carcara, if the reviewer finds it better.


carcara/src/translation/eunoia/ast.rs
bpandreotti, line 1: I think this module should just be called `ast`, since it's already inside the `eunoia` module.

carcara/src/translation/eunoia/eunoia_ast.rs
```diff
@@ -0,0 +1,338 @@
+//! AST representation of a fragment of Eunoia required to mechanize Alethe proofs.
+use crate::translation::Symbol;
+
+pub struct EunoiaTheorySignature;
```
bpandreotti: It looks like this struct is not used anywhere. Can it be removed, or will it be used in a future PR?

carcara/src/translation/eunoia/eunoia_ast.rs
```diff
@@ -0,0 +1,338 @@
...
+                panic!()
+            }
+        }
+    }
+}
+
+/// A collection of proof rules.
+pub struct EunoiaProofRules {}
+
+pub struct EunoiaProgram;
```
bpandreotti: Similarly to `EunoiaTheorySignature`, will these be used later?

carcara/src/translation/eunoia/eunoia_ast.rs
```diff
@@ -0,0 +1,338 @@
...
+    // is applied to.
+    // Internally, (! T :implicit) drops T from the list of arguments of the function
+    // type we are defining.
+
+    // TODO: it looks that there are these KindParam, used to define variables that
+    // refer to types that inhabit Type (or "kinds"?); while there is also a
+    // "typed-param" syntactic category that refers to the parameters of a function:
+    // values inhabiting some given type
+    KindParam(EunoiaType, Vec<EunoiaTypeAttr>),
+}
```
bpandreotti: It's not super clear to me from the comments, but is the only type of `KindParam` a series of `EunoiaTypeAttr`s? In other words, is `KindParam` supposed to be the only variant of the enum? If so, this is more clear as a tuple struct:
```rust
pub struct EunoiaKindParam(EunoiaType, Vec<EunoiaTypeAttr>);
```

carcara/src/translation/eunoia/eunoia_ast.rs
```diff
@@ -0,0 +1,338 @@
...
+    // <decimal> denoting the category of decimals -?<digit>+.<digit>+,
+    Decimal(rug::Rational),
+
+    // <rational> denoting the category of rationals -?<digit>+/<digit>+,
+    Rational(rug::Integer, rug::Integer),
+
+    // <string> denoting the category of string literals "<char>*"
+    String(String),
+    // TODO: is it reasonable/required a specific syntactic category "constant"?
+    // Like Const(Constant) and then define enum Constant?
```
bpandreotti: In my opinion, it's better to not have a separate `Constant` struct if you can avoid it. For the Alethe AST we end up needing to work with constants separately from terms often enough that a `Constant` struct is necessary, but I would prefer to keep things simple for the Eunoia AST. If possible, of course.

carcara/src/translation/eunoia/ast.rs
```diff
@@ -0,0 +1,338 @@
...
+    // To capture the situation where a list of terms are
+    // actually an evaluation of some given function over
+    // actual parameters.
+    App(Symbol, Vec<EunoiaTerm>),
+
+    // Explicit higher-order function application ("_" symbol)
+    HOApp(Box<EunoiaTerm>, Vec<EunoiaTerm>),
+
+    // Application of a built-in operator
+    Op(EunoiaOperator, Vec<EunoiaTerm>),
```
bpandreotti, line 167: I see you are referring to terms using `Box<_>` and `Vec<_>` directly, i.e., without hash consing (and using `Rc<_>`) like we do for Alethe. This means that constructing a term `(+ x x)` needs to allocate the term `x` twice, which can maybe lead to performance issues.

Of course, it would be pretty annoying to have to implement an entire separate `TermPool` just for Euonia terms, but since we are translating _from_ Alethe, maybe we can abuse the fact that the existing Alethe ast is already "deduplicated".

What I mean is: when translating an Alethe term `t` into an Eunoia term `t'`, you check if you already translated it and just clone the existing `Rc<EunoiaTerm>` representing `t'`. You would need to keep a cache of translated terms, something like `HashMap<Rc<Term>, Rc<EunoiaTerm>>`. This is assuming that a term `t` always becomes the same term `t'` when translating, regardless of the context around it, which I'm not sure is true for the Eunoia translation.

In any case, we shouldn't worry about this for now before making sure this is actually a performance bottleneck. Still, it's good to be aware.

carcara/src/translation/eunoia/eunoia_ast.rs
```diff
@@ -0,0 +1,338 @@
...
+#[derive(Debug, PartialEq)]
+pub enum EunoiaCommand {
+    // Eunoia commands
+
+    // TODO:
+    // The command:
+    // (assume s f)
+    // can be seen as syntax sugar for:
+    // (declare-const s (Proof f))
+    // how to deal with these syntax sugars?
```
bpandreotti: Since we are _producing_ Eunoia code, I think there are two reasonable ways to deal with syntax sugar:

1. Actually represent a separate `assume` command in the AST, as you've done for now.
2. Store a "desugared" AST where you only have a `declare-const` command and no `assume`. Then, only when printing the proof do we detect the cases where `declare-const` could be replaced by `assume` (that is, replace `(declare-const s (Proof f))` by `(assume s f)` whenever possible).

While (2) would simplify the AST and may be simpler to work with, I lean towards (1), since it is more "faithful" to the intent of the proof. What do you think?

carcara/src/translation/eunoia/alethe_signature/theory.rs
```diff
@@ -0,0 +1,201 @@
...
+    // Binders.
+    pub let_binder: Symbol,
+    pub forall_binder: Symbol,
+    pub exists_binder: Symbol,
+    pub choice_binder: Symbol,
+
+    // VarList constructors
+    pub varlist_cons: Symbol,
+    pub varlist_nil: Symbol,
+}
```
bpandreotti, line 63: Since (I assume) these fields will never change, consider making them `&'static str` instead of `String`s. This avoids allocating for each of them and is generally simpler to use.

bpandreotti (commented):
Some initial comments on the Eunoia AST and signature. I will do a later review covering the Eunoia printer and the TSTP module as a whole.

carcara/src/translation/eunoia/ast.rs
HanielB: The best way to track this is by opening an issue, so it does not get lost. 

In cvc5's internal translation between proof calculi this question also arises. There the solution has been to piggyback on the equivalent of `Term` to represent the external terms, which has the advantage that you can enjoy the existing hash-consing but the disadvantage that you lose dedicated checks for the external terms being well-formed. Maybe the best approach is indeed to do what Bruno suggested of keeping an extra map between the two representations so you can have the best of both worlds.

carcara/src/translation/eunoia/ast.rs
Mallku2: You are right! Done.

carcara/src/translation/eunoia/eunoia_ast.rs
Mallku2: You are right, it was planeed at some point during the beginning of the project, but it ended up being discarded. I removed it.

carcara/src/translation/eunoia/eunoia_ast.rs
Mallku2: Ups, another one. They won't be used. I removed them.

carcara/src/translation/eunoia/eunoia_ast.rs
Mallku2: Yes, you are right. I think that at the beginning I was planning to cover more structure within this single enum, but I ended up considering just that single constructor. I am changing that right now, and sorry for the messy stuff!

carcara/src/translation/eunoia/eunoia_ast.rs
Mallku2: That is really useful feedback, because, indeed, we do not have many different kind of constants literals in Eunoia. I will let it as it is, and also remove those TODOs!

carcara/src/translation/eunoia/ast.rs
Mallku2: This is a sensible recommendation, and one that I totally neglected from the beginning of the project. Right now I have 500 certificates that cannot be translated under current constraints of time and/or memory, just from the QF_UF benchmark set. There is not even such kind of fundamental optimization implemented, yet. I will open an issue to keep track of this, while studying the current timeouts/memouts.

carcara/src/translation/eunoia/eunoia_ast.rs
Mallku2: Yes, I am also leaning towards the first option. Indeed, desugaring the command might obscure the connection between the original Alethe proof and its Eunoia counterpart.

carcara/src/translation/eunoia/alethe_signature/theory.rs
Mallku2: Indeed! I already changed it, following your suggestion.

carcara/src/translation/eunoia/printer.rs
```diff
@@ -0,0 +1,496 @@
...
+            write!(self.sink, "{}", tag)?;
+        } else {
+            // {not args.is_empty()}
+            // S-expression has the form (tag arg1 ...)
+            write!(self.sink, "(")?;
+            write!(self.sink, "{}", tag)?;
+            // TODO: how to propagate errors from within the lambda abstraction?
+            args.iter().for_each(|arg| {
+                let _ = write!(self.sink, " {}", arg);
+            });
```
bpandreotti: Two options to do that. Use [`try_for_each`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each):
```suggestion
            args.iter()
                .try_for_each(|arg| write!(self.sink, " {}", arg))?;
```
or just use a for loop:

```suggestion
            for arg in args {
                write!(self.sink, " {}", arg)?;
            }
```

carcara/src/translation/tstp/printer.rs
```diff
@@ -0,0 +1,688 @@
...
+                parents.iter().for_each(|parent| {
+                    if first_iteration {
+                        ret += &TstpPrinter::source_to_concrete_syntax(parent);
+                        first_iteration = false;
+                    } else {
+                        // { ! first_iteration }
+                        ret += ", ";
+                        ret += &TstpPrinter::source_to_concrete_syntax(parent);
+                    }
+                });
```
bpandreotti: One simpler way would be to do this:
```suggestion
                ret += &TstpPrinter::source_to_concrete_syntax(&parents[0]);
                for parent in &parents[1..] {
                    ret += ", ";
                    ret += &TstpPrinter::source_to_concrete_syntax(parent);
                }
```
I would recommend putting that logic in a separate function, since it's also useful in some other places. Something like:
```rust
fn print_sequence<T>(seq: &[T], func: fn(&T) -> String) -> String {
    let mut result = func(&seq[0]);
    for item in &seq[1..] {
        result += ", ";
        result += &func(item);
    }
    result
}
```

carcara/src/translation/tstp/printer.rs
bpandreotti, line 1: One general note about how you are doing the printers: Instead of having a function for each type that converts it into a `String`, you could instead implement the `std::fmt::Display` trait for each of these types. This would avoid allocating strings everywhere (so it might be more performant) and it would make the code more organized (instead of having a bunch of `foo_bar_to_concrete_syntax` functions you can just call `format!("{}", foo_bar)`)

There are downsides to this, however. For one, it would be a major change and you would need to pretty much rewrite the printer modules. More importantly, if I recall correctly, there is some weird stuff related to `std::io::Write` and `std::fmt::Write` being different, incompatible traits, which I ran into when writing the Alethe pretty printer. Since you're using `io::Write` in this module, and `fmt::Display` works with `fmt::Write`, things might get tricky.

I think it's best to keep the current strategy for now, but this might be a good idea for a future refactor.

carcara/src/translation/tstp/ast.rs
```diff
@@ -0,0 +1,354 @@
...
+    ) -> Self {
+        Self {
+            language: provided_language,
+            name: provided_name,
+            role: provided_role,
+            formula: provided_formula,
+            source: provided_source,
+            useful_info: provided_useful_info,
+        }
+    }
```
bpandreotti, line 41: Why the `provided_` prefix? You can just use the field names directly, it even simplifies the struct initialization:
```suggestion
    pub fn new(
        language: TstpLanguage,
        name: Symbol,
        role: TstpFormulaRole,
        formula: TstpFormula,
        source: TstpAnnotatedFormulaSource,
        useful_info: Symbol,
    ) -> Self {
        Self {
            language,
            name,
            role,
            formula,
            source,
            useful_info,
        }
    }

```

carcara/src/translation/tstp/printer.rs
HanielB: > I think it's best to keep the current strategy for now, but this might be a good idea for a future refactor.

Agreed. Again, best to open an issue about this.

carcara/src/translation/eunoia/printer.rs
Mallku2: Nice!, many thanks, I will be using the second suggestion. Done!

carcara/src/translation/tstp/printer.rs
Mallku2: Many thanks for this!, I have implemented the separate function.

carcara/src/translation/tstp/printer.rs
Mallku2: Thanks for this suggestion!, I will implement it once we finish merging the current iteration of the compiler. In the meantime, I will open an issue, as suggested.

carcara/src/translation/tstp/ast.rs
Mallku2: Can't find comments in this regard in my notes xD I implemented your suggestion!

carcara/src/translation/tstp/printer.rs
bpandreotti: Awesome, thanks!

bpandreotti (approved):
Thanks for all the changes, the code looks good. Are there any TODOs left for this PR, or can I merge?

Mallku2:
> Thanks for all the changes, the code looks good. Are there any TODOs left for this PR, or can I merge?

You can merge!

PR #135, opened 2026-08-08
Alethe2Eunoia and Alethe2Tstp compilation routines

Mallku2:
This PR presents the actual compilation routines for both compilers. Some observations worth mentioning:
- Some minor changes were made in file `carcara/src/translation/eunoia/alethe_signature/theory.rs`, as part of an effort to fix some pending TODOs and refactor a little bit the code of Alethe2Eunoia.
- There are still some pending TODOs in Alethe2Eunoia which mostly refer to parts of Alethe not yet translated into Eunoia (e.g., bit-vectors), or related doubts. If there is a policy against have such TODOs in source code, I can just remove them from this PR.
- Alethe2Tstp is in a rather immature state, since we lack a serious benchmark set to put it to stress. Hence, its source code is more experimental. If it is preferable for the reviewer to skip its review for the time being, I can also remove it from the PR, until the code is more mature.


carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+use crate::ast::*;
+use crate::translation::eunoia::alethe_signature::theory::*;
+use crate::translation::eunoia::ast::*;
+use crate::translation::Symbol;
+use crate::translation::Translator;
+use crate::translation::TranslatorData;
+use crate::translation::VecToVecTranslator;
+
+// Deref for ast::rc::Rc<Term>
+use std::ops::Deref;
```
bpandreotti: Instead of using `.deref()` to get a `&Term` from a `&Rc<Term>`, I would prefer if you used `.as_ref()`. It's what is mostly used in the rest of Carcara, and doesn't require this trait import.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
+//! Translator for `EunoiaProof`.
+use crate::ast::*;
+use crate::translation::eunoia::alethe_signature::theory::*;
+use crate::translation::eunoia::ast::*;
+use crate::translation::Symbol;
+use crate::translation::Translator;
+use crate::translation::TranslatorData;
+use crate::translation::VecToVecTranslator;
```
bpandreotti: This is very nitpicky, but you can merge these imports so they're more compact:
```suggestion
use crate::translation::{
    eunoia::{alethe_signature::theory::*, ast::*},
    Symbol, Translator, TranslatorData, VecToVecTranslator,
};
```

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+                id: id.to_owned(),
+                conclusion_clause: Some(conclusion),
+                rule: rule.clone(),
+                premises: EunoiaList { list: premises },
+                arguments: EunoiaList { list: arguments },
+            });
+    }
+}
+
+impl VecToVecTranslator<'_, EunoiaCommand, EunoiaTerm, EunoiaType, Symbol> for EunoiaTranslator {
```
bpandreotti: I probably should have thought of this when reviewing the previous PR, but I think this trait should have these types as [associated types](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html) instead of generic type paramters.

That is, you would declare the trait in `translation/mod.rs` like this:
```rust
pub trait VecToVecTranslator<'a> {
    type StepType;
    type TermType: Clone + 'a;
    type TypeTermType: Clone + 'a;
    type OperatorType;

    // ...
}
```

Implement it here like this:
```rust
impl VecToVecTranslator<'_> for EunoiaTranslator {
    type StepType = EunoiaCommand;
    type TermType = EunoiaTerm;
    type TypeTermType = EunoiaType;
    type OperatorType = Symbol;
    // ...
}
```

And generally use `Self::StepType`, etc instead of just `StepType`.

The semantic difference between those is that with generics, you can have more than one implementation for the same type, using different type parameters:
```rust
impl VecToVecTranslator<'_, Foo, ...> for EunoiaTranslator { ... }

impl VecToVecTranslator<'_, Bar, ...> for EunoiaTranslator { ... }
```
whereas with associated types you can have only one implementation, which sets which types are used. For our use case, I think the latter is more appropriate. What do you think?

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+            Term::Var(string, _) => {
+                // Check if it is a variable introduced by some binder
+                match self
+                    .get_read_translator_data()
+                    .alethe_scopes
+                    .get_variable_in_scope(string)
+                {
+                    Some(_) => self.build_var_binding(string),
+
+                    None => EunoiaTerm::Id(string.clone()),
```
bpandreotti, line 398: Does this `None` case happen in practice? I would expect Carcara's parser to ensure that all `Term::Var` terms map to a variable defined in an outer scope.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+                final_let_trans
+            }
+
+            Term::Binder(binder, binding_list, scope) => {
+                // New scope to shadow those context variables that
+                // now bound by this binder.
+                // TODO: reusing variables_in_scope concept
+                // for this new kind of scope (not the one
+                // related with contexts introduced through
+                // "anchor" commands).
```
bpandreotti: I think it is better to indeed use `variable_in_scope` for both kinds of scopes, binders and anchors. That mirrors how we use the symbol table in the Alethe parser.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+            binding_occ.push(EunoiaTerm::Var(
+                name.clone(),
+                Box::new(EunoiaTerm::Type(translated_value_sort)),
+            ));
+
+            values.push(translated_value.clone());
+        });
+
+        (EunoiaTerm::List(binding_occ), values)
+    }
```
bpandreotti, line 599: Instead of wrapping the first list in `EunoiaTerm::List`, this function could return the `Vec<EunoiaTerm>` directly, so you don't have to unwrap it at the call site.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+
+                        _ => {
+                            panic!();
+                        }
+                    },
+
+                    None => {
+                        panic!();
+                    }
+                };
```
bpandreotti: Instead of matching and panicking on the `None` case, it's simpler to just call `.unwrap()`:
```suggestion
                let term = sorts.last().unwrap();
                match term.as_ref() {
                    Term::Sort(sort) => {
                        return_sort = EunoiaTranslator::translate_sort(sort);
                    }
                    _ => panic!(),
                }
```

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+                    "let" => {
+                        // Include, as premises, previous step from the actual subproof.
+                        match previous_command_id {
+                            Some(id) => eunoia_premises.push(EunoiaTerm::Id(id.to_owned())),
+
+                            None => {
+                                println!("'let' step without premises?");
+                                panic!();
+                            }
+                        }
```
bpandreotti: Again, it's better to use `unwrap` here (or `expect`, if you want to have a custom panic message)
```suggestion
                        eunoia_premises.push(EunoiaTerm::Id(
                            previous_command_id
                                .expect("'let' step without premises?")
                                .to_owned(),
                        ));
```
The same applies to the code for some other rules further down.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1112 @@
...
+                            .push(EunoiaTerm::Id(self.get_last_introduced_context_id()));
+
+                        self.translate_generic_step_pop(
+                            id,
+                            conclusion,
+                            rule,
+                            eunoia_premises,
+                            eunoia_arguments,
+                        );
+                    }
```
bpandreotti: I think you might be able to merge some of these `match` arms. The code for `bind_let` here is identical to the one for `bind` and `sko_ex`.

bpandreotti (commented):
For now, some comments on Alethe2Eunoia. I'll review Alethe2TSTP later, but I think it should be ok to keep it on this same PR.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: I see!, done.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: You are right. I don't even know why I left things that way. Done!

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: As suggested in the article that you shared, I see the benefit in syntax and I have no problem with changing that. I am a bit confused with the benefits in semantics, though, but will review this more in depth tomorrow and update the code as you suggested!

Mallku2: Now I understand the benefits, and I have implemented the changes following your suggestions. Many thanks for the observation!

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Carcara represents as `Term::Var` both, variables bound by some binder and, say, identifiers representing undefined constants from the input SMT problem. Am I right? In our representation in Eunoia we need to distinguish both cases, so we need to reason about variables in scope to do so. We keep track of binding occurrences of variables, to distinguish bound variables from identifiers representing problem constants.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Nice!, removing that TODO and feeling relief.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Nice catch!, I don't even know what I was trying to achieve over there xD Changed that!

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Done!

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Changed every place where I was not taking advantage of this, and also a little embarrassed for not been using this before.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Indeed! I started doing that before this PR, but there are still some opportunities for refactoring. I have implemented that now.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
bpandreotti: Ah, that makes sense, thanks for the clarification.

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1012 @@
...
+                    eunoia_arguments.push(self.translate_term(arg));
+                });
+
+                match rule.as_str() {
+                    // Subproof-closing steps
+                    "let" | "bind_let" | "bind" | "sko_ex" => {
+                        self.translate_generic_step_pop(
+                            id,
+                            conclusion,
+                            "let_elim",
```
bpandreotti: Is this right? You're mapping all of those rules to `let_elim`. Based on the preivous version of the code, I assume you wanted to this:
```suggestion
                            if rule == "let" { "let_elim" } else { rule },
```

carcara/src/translation/eunoia/alethe_2_eunoia.rs
```diff
@@ -0,0 +1,1012 @@
...
+                        let mut cl_disjuncts: Vec<EunoiaTerm> = vec![];
+
+                        // Id of the premise step
+                        let mut id_premise: Symbol = "".to_owned();
+
+                        // TODO: some more efficient way to deal with
+                        // the fact that we use a "stack" of assumptions?
+                        let mut discharge_copy = discharge.clone();
+                        discharge_copy.reverse();
+                        discharge_copy.iter().for_each(|discharged_assumption| {
```
bpandreotti: You can avoid cloning here by calling `.rev()` on the iterator:
```suggestion
                        discharge.iter().rev().for_each(|discharged_assumption| {
```

bpandreotti (commented):
A few more comments on Alethe2Eunoia

carcara/src/translation/tstp/alethe_2_tstp.rs
```diff
@@ -0,0 +1,1092 @@
...
+        match role {
+            TstpFormulaRole::Assumption => {
+                // TODO: for the moment we are just using Alethe step's id.
+                // new_name = "assumption".to_owned();
+                new_name = step_id.to_owned();
+            }
+
+            TstpFormulaRole::Axiom => {
+                new_name = "axiom".to_owned();
+                new_name += &("_".to_owned() + &self.formulas_count[role].to_string());
```
bpandreotti: I think it's clearer to write this as:
```suggestion
                new_name += &format!("axiom_{}", self.formulas_count[role]);
```
Same applies to the `TstpFormulaRole::Type` case further down.

carcara/src/translation/tstp/alethe_2_tstp.rs
```diff
@@ -0,0 +1,1092 @@
...
+
+        annotated_formula
+    }
+
+    /// Abstracts the mechanism used to generate a new name for an annotated
+    /// formula, and register it, for future reference.
+    fn new_formula_name(&mut self, step_id: &str, role: &TstpFormulaRole) -> String {
+        // For axioms, lemmas, conjecture, hypothesis, naming
+        // mechanism is rather simple: role + number.
+        let mut new_name: String;
```
bpandreotti: More generally, instead of declaring this `new_name` variable, assigning to it in the `match` expression and then returning it, you can just return the `match` expression directly.

carcara/src/translation/tstp/alethe_2_tstp.rs
```diff
@@ -0,0 +1,1092 @@
...
+                            ret = TstpFormula::BinaryOperatorApp(
+                                binary_op.clone(),
+                                Box::new(ret.clone()),
+                                Box::new(operand.clone()),
+                            );
+                        } else {
+                            // { position <= 1 }
+                            position += 1;
+                        }
+                    });
```
bpandreotti: You can use [`Iterator::skip`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip):

```suggestion
                    operands_tstp.iter().skip(2).for_each(|operand| {
                        ret = TstpFormula::BinaryOperatorApp(
                            binary_op.clone(),
                            Box::new(ret.clone()),
                            Box::new(operand.clone()),
                        );
                    });
```

bpandreotti: Thinking about it more, you can do this more succintly with a `.reduce()` call:
```suggestion
                    operands_tstp.into_iter().reduce(|a, b| {
                        TstpFormula::BinaryOperatorApp(binary_op.clone(), Box::new(a), Box::new(b))
                    });
```
Reduce goes left to right, so this is left associative.

bpandreotti (changes requested):
Initial comments on Alethe2TSTP

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: What a silly mistake from my part xD Sorry!, I am removing that and just translating "let" as "let" (and, also, adding some tests in my test suite, to catch such stuff). This is an old artifact of the first iterations of the mechanization, I think because "let" was a reserved word in Eunoia back then. Fixed, thanks!

carcara/src/translation/eunoia/alethe_2_eunoia.rs
Mallku2: Super!, done.

carcara/src/translation/tstp/alethe_2_tstp.rs
Mallku2: Absolutely loving using string interpolation. Changed that and every other place where I was doing string concatenation.

carcara/src/translation/tstp/alethe_2_tstp.rs
Mallku2: Indeed!, was using that approach since I was doing string concatenation, but now, with string interpolation, it is useless to do it so. Changing it.

carcara/src/translation/tstp/alethe_2_tstp.rs
Mallku2: Super!, done.

Mallku2: Really nice!, done.

carcara/src/translation/tstp/alethe_2_tstp.rs
```diff
@@ -0,0 +1,1067 @@
...
+            Operator::GreaterThan => TstpOperator::BinaryOperator(TstpBinaryOperator::Greater),
+
+            Operator::GreaterEq => TstpOperator::BinaryOperator(TstpBinaryOperator::GreaterEq),
+
+            Operator::LessThan => TstpOperator::BinaryOperator(TstpBinaryOperator::Less),
+
+            Operator::LessEq => TstpOperator::BinaryOperator(TstpBinaryOperator::LessEq),
+
+            // TODO:?
+            Operator::Distinct => TstpOperator::BinaryOperator(TstpBinaryOperator::Inequality),
```
bpandreotti: In `translate_operator_application`, we are assuming that all `TstpBinaryOperator`s are left-associative, but in Alethe some of these are not---namely `=>` is right-associative, and `=` and `distinct` are chainable---so you are mistranslating them here. I think the best way to handle this is to check the operator associativity in `translate_operator_application`.

carcara/src/translation/tstp/alethe_2_tstp.rs
```diff
@@ -0,0 +1,1067 @@
...
+                    // {!clause.is_empty()}
+                    self.translate_operator_application(Operator::Or, clause)
+                };
+
+                // NOTE: not adding conclusion clause to this list
+                let mut tstp_arguments: Vec<TstpFormula> = Vec::new();
+
+                args.iter().for_each(|arg| {
+                    tstp_arguments.push(self.translate_term(arg));
+                });
```
bpandreotti: You are translating the arguments twice. You already translated them into the  `alethe_translated_args` variables, so you can remove this.
```suggestion
```

bpandreotti (commented):
A few more comments on Alethe2TSTP

bpandreotti:
There are still many TODOs for the Alethe2TSTP module. I think it might be better after all to split that into a separate PR, so we can merge the Eunoia part of this. What do you think?

carcara/src/translation/tstp/alethe_2_tstp.rs
Mallku2: Sorry, I am confused here (long time not seeing that part of the code, I think): 
- at the level of ASTs we do not have "associativity", right?
- Aren't assoc. properties only considered for concrete syntax, and applied when generating the corresponding AST? That is, if one writes in an Alethe certificate "a => b => c", the AST generated for that will represent something like "a => (b => c)". While, in an TSTP certificate the corresponding AST would represent something like "(a => b) => c". I think that, beyond the comments that I put in the source code, the assumption during translation is that associativity is already solved by Carcara's parser.

carcara/src/translation/tstp/alethe_2_tstp.rs
Mallku2: You are right!

Mallku2:
> There are still many TODOs for the Alethe2TSTP module. I think it might be better after all to split that into a separate PR, so we can merge the Eunoia part of this. What do you think?

Yes, indeed. I removed the module, and will open later a PR just for it, as a "work in progress" thing not to be reviewed until some further testing and polishing occurs.

bpandreotti (approved):
Thanks for all the changes! Are there any TODOs left for Alethe2Eunoia, or can we merge?

Mallku2:
> Thanks for all the changes! Are there any TODOs left for Alethe2Eunoia, or can we merge?

There are some minor TODOs in the code, but that are intended for future improvements. No need to discuss them at the moment. I have pushed just now a small change related with how I pass contexts to the `refl` rule (intended to harmonize this with I do the same for other rules), and reverted a change related with how I translate the name of the rule `let`. I am just mentioning this for completeness, there are no important changes in the commit that would need further review. We can merge. 

EDIT: fixed the CI errors, we can merge!

bpandreotti (approved):

PR #136, opened 2026-08-10
Check that terms with the same name in match_term! are equal

netolcc06:
Implements the second half of #118. The first half (flattening the macro so patterns of arbitrary depth work) was done in #122


bpandreotti (approved):
Awesome! It's great how this makes checking some rules trivial.

PR #137, opened 2026-08-11
Add automata infrastructure and AST term integration

vinisilvag:
This PR introduces finite automata infrastructure into Carcara to support proof checking and elaboration of regular constraint reasoning in Alethe proofs. This work is motivated by recent developments in string-solving SMT tools such as OSTRICH, which is being extended to generate proofs in the Alethe format. The theoretical background for decision procedures and constraint propagation over regular languages follows the foundations presented in

> ["_The Power of Regular Constraint Propagation_"](https://dl.acm.org/doi/10.1145/3763165)
> (Matthew Hague, Artur Jeź, Anthony Widjaja Lin, Oliver Markgraf, Phlipp Rümmer, 2025).

Rather than relying on external third-party automata libraries, implementing a dedicated finite automata module directly in Carcara provides full flexibility and fine-grained control over internal representations, customized algorithms, transition models, and proof verification operations.

Overview of key additions:
  * Core Automata Infrastructure: Introduced `Automaton`, `State`, `Transition`, and `Trigger` types supporting both NFAs and DFAs with symbol range triggers and epsilon transitions, alongside a Disjoint Set Union (DSU) structure for equivalence testing.
  * Automata Operations: Implemented Hopcroft-Karp DFA equivalence checking, syntactic subautomaton verification, product-construction intersection, totalization, complementation, BFS-based reachability, and subset-construction determinization.
  * AST and Parser Integration: Added AST support for regular language constants (`RegLan`), parser combinators for reading automata definitions, error handling variants, and direct conversion of SMT-LIB regular expression operators (`re.++`, `re.*`, `re.union`, `re.inter`, `str.to_re`, etc.) into automata.
  * Testing and Documentation: Added unit test coverage across all automata operations, algorithms, and regex conversions, along with rustdoc documentation.

Future PRs will add verification procedures for the proof rules and optimizations to handle automata more efficiently.


carcara/src/checker/error.rs
```diff
@@ -240,6 +241,32 @@ pub enum CheckerError {
...
+    },
+
+    #[error("regular expression replace failed: replacing in '{s}' with regex '{regex}' and replacement '{replacement}' expected to result in '{expected}', got '{got}'")]
+    RegexReplaceFailed {
+        s: String,
+        regex: Rc<Term>,
+        replacement: String,
+        expected: String,
+        got: String,
+    },
```
bpandreotti: Clippy is complaining that the `CheckerError` enum is too large (which is not great for performance), and I think it's because of this variant specifically (each `String` is 24 bytes, `Rc<Term>` is 8 bytes, so that's 104 bytes at least; plus the discriminant and possibly padding).

I think the best way to solve this is to move this and other related error variants to a separate enum, maybe something like `StringError` (like we have `QuantiferError`, `CongruenceError`, etc). Then, in `CheckerError`, you add a variant:
```rust
#[error(transparent)]
String(#[from] Box<StringError>),
```
The `Box` indirection makes this variant take only 8 bytes, reducing the size of `CheckerError`. The `#[error(transparent)]` attribute is needed to forward the proper error messages from the inner enum.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1354 @@
...
+                    // complementation genuinely requires a (complete) DFA
+                    let dfa = if a.is_nfa() {
+                        Automaton::determinize(&a)
+                    } else {
+                        a
+                    };
+                    Ok(dfa.complement())
+                }
+                Term::Op(Operator::ReRange, args) => {
+                    let c1_term = args.get(0).ok_or(CheckerError::Unspecified)?;
```
bpandreotti: Clippy also complains about using `.get(0)` here instad of `.first()`:
```suggestion
                    let c1_term = args.first().ok_or(CheckerError::Unspecified)?;
```
A bit silly.

Anyhow, the larger problem in my view is that you're using `CheckerError::Unspecified` in some places. Usually, we want to have dedicated error types, with nicer error messages.

In this case, though, I think these invariants you're relying on are all guaranteed by the parser (namely, the number and type of arguments for `re.range`), which means you can panic if they're broken, since that would mean a bug in the parser. So, I think we can just do this: 
```suggestion
                    let c1_term = &args[0];
```

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1354 @@
...
+                    let c2_term = args.get(1).ok_or(CheckerError::Unspecified)?;
+                    let Term::Const(Constant::String(s1)) = c1_term.as_ref() else {
+                        return Err(CheckerError::Unspecified);
+                    };
+                    let Term::Const(Constant::String(s2)) = c2_term.as_ref() else {
+                        return Err(CheckerError::Unspecified);
+                    };
```
bpandreotti: Following up on the previous comment, we can panic here instead of returning an error. I personally like to use `unreachable!()` for these cases.
```suggestion
                    let Term::Const(Constant::String(s1)) = c1_term.as_ref() else {
                        unreachable!()
                    };
                    let Term::Const(Constant::String(s2)) = c2_term.as_ref() else {
                        unreachable!()
                    };
```

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1354 @@
...
+                    let c1 = s1.chars().next().ok_or(CheckerError::Unspecified)? as u16;
+                    let c2 = s2.chars().next().ok_or(CheckerError::Unspecified)? as u16;
```
bpandreotti: This invariant (the strings being non-empty), however, is not guaranteed by the parser. In fact, according to the [SMT-LIB Strings theory](https://smt-lib.org/theories-UnicodeStrings.shtml), giving an empty string to `re.range` is allowed, and it evaluates to the empty language.

You can implement a dedicated error for this case, but it would be even better to correctly parse this, and produce an automaton corresponding to the empty language.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1354 @@
...
+                                    transitions: HashSet::new(),
+                                }],
+                                initial_state: 0,
+                            });
+                        }
+                        automatons.push(a);
+                    }
+
+                    let mut automatons = automatons.into_iter();
+                    let mut result = automatons.next().ok_or(CheckerError::Unspecified)?;
```
bpandreotti: `re.++` is guaranteed to take at least one argument, so you can safely unwrap here instead of using `CheckerError::Unspecified`.
```suggestion
                    let mut result = automatons.next().unwrap();
```

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1354 @@
...
+                    // the product construction works directly on NFAs once
+                    // their epsilon transitions are eliminated, so the
+                    // components need not be determinized
+                    let mut components = Vec::new();
+                    for re in inter {
+                        let nfa = rec_create_from_regex_operators(pool, re)?;
+                        components.push(nfa.epsilon_eliminate());
+                    }
+                    let mut components = components.into_iter();
+                    let mut res = components.next().ok_or(CheckerError::Unspecified)?;
```
bpandreotti: Same thing here, `re.inter` will always have at least one argument.

carcara/src/parser/mod.rs
```diff
@@ -242,6 +243,18 @@ impl<'p, 's> Parser<'p, 's> {
...
+        match parse_automaton(automaton_repr.trim()) {
+            Ok((remaining, automata)) => {
+                if !remaining.is_empty() {
+                    return Err(ParserError::InvalidAutomatonDeclaration(automaton_repr));
+                }
+                Ok(automata)
+            }
+            Err(_) => Err(ParserError::InvalidAutomatonDeclaration(automaton_repr)),
+        }
+    }
```
bpandreotti: This method doesn't use the `self` parameter, so it could just be a free function. However, since it's relatively short and is only used in a single place, I wouldn't have it as a separate method in the first place, and would just inline this code in the call site.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1354 @@
...
+                        id: "init".to_owned(),
+                        accept: false,
+                        transitions: HashSet::from_iter(
+                            [Transition {
+                                to: 0,
+                                trigger: Trigger::Range((0, u16::MAX)),
+                            }]
+                            .iter()
+                            .cloned(),
+                        ),
```
bpandreotti: Clippy is complaining that using `from_iter` here is not idiomatic, and suggests calling `.collect()` on the iterator. Even better, I think, would be to use `HashSet::from`, since `HashSet<T>` implements `From<[T; N]>`:
```suggestion
                        transitions: HashSet::from([Transition {
                            to: 0,
                            trigger: Trigger::Range((0, u16::MAX)),
                        }]),
```

bpandreotti (changes requested):
Some initial suggestions to fix the linter warnings. I'll do a more thorough review later.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+
+impl Hash for State {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let mut transitions_vec: Vec<_> = self.transitions.iter().collect();
+        transitions_vec.sort_by(|a, b| a.to.cmp(&b.to).then_with(|| a.trigger.cmp(&b.trigger)));
+        for transition in transitions_vec {
+            transition.hash(state);
+        }
+    }
+}
```
bpandreotti: I see you are sorting here to make sure the hash is deterministic. I would prefer instead to use a `BTreeSet` for the transitions, so we can just `#[derive(Hash)]`. This avoids having to to this vec allocation and sort, and avoids having a manual implementation of `Hash` (which can lead to bugs if someone adds a field to `State` and forgets to hash it here, for example).

The performance of `BTreeSet` is in theory worse than `HashSet` (O(log n) vs O(1)), but I would bet that in practice the difference is minor and avoiding this allocation and sort when hashing will probably have a bigger impact.

Alternatively, `IndexSet` is a hash set that has a stable order when iterating. If it is acceptable that two automata with the same transitions (but inserted in different orders) have different hashes, we could use that instead. I still lean towards `BTreeSet`, though. What do you think?

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+impl Automaton {
+    // Construct an automaton based on the automaton name, initial state id, the complete set of
+    // transitions (from, to, range), and the set of accepting states
+    fn new(
+        automaton_name: &str,
+        initial_state_id: &str,
+        transitions: Vec<(&str, &str, (u16, u16))>,
+        accepting_states: Vec<&str>,
+    ) -> Automaton {
+        let mut accepting_states_map = HashSet::new();
```
bpandreotti: This should be name `accepting_states_set`, since it's a set. :)

Or just `accepting_states`, you can shadow the parameter name.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+                for (index, state) in all_states.iter().enumerate() {
+                    if state.id == *id.to_owned() {
+                        found = Some(index);
+                        transition_ids.push(index);
+                    }
+                }
+                if found.is_none() {
+                    all_states.push(State::new(id, accepting_states_map.contains(id)));
+                    transition_ids.push(all_states.len() - 1);
+                }
```
bpandreotti: You can use the built-in [`find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) method here:
```suggestion
                match all_states.iter().enumerate().find(|(_, s)| s.id == *id) {
                    Some((index, _)) => transition_ids.push(index),
                    None => {
                        all_states.push(State::new(id, accepting_states_map.contains(id)));
                        transition_ids.push(all_states.len() - 1);
                    }
                }
```
Also, note that for each transition in each new state, you are traversing the vec of states. It would be better to keep a `HashMap<&str, StateId>`, mapping state names to their index in the vec. It would also allow you to detect two states that have the same name (which I assume must be invalid).

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+                        ranges.push(range);
+                    }
+                    Trigger::Epsilon => {
+                        return true;
+                    }
+                };
+            }
+            // NFA if the automaton has overlapping ranges on transitions outgoing the same state
+            // (non-determinism)
+            if has_overlapping_ranges(ranges.clone()) {
```
bpandreotti, line 216: This will also be made simpler if we use a `BTreeSet` for the transitions, since they will already be sorted.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+                return true;
+            }
+            // NFA if the automaton does not have transitions for every range from 0 to u16::MAX
+            if !missing_ranges(&ranges, 0, u16::MAX).is_empty() {
```
bpandreotti: I'm surprised that you are using `u16` for the characters---there are a lot more than 2^16 characters in unicode! The Rust `char` type, for example, is represented with a `u32`. Is the automaton format specified such that the characters will always be limited to those less than `u16::MAX`?

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+            for transition in &state.transitions {
+                transitions.push((
+                    state.clone(),
+                    self.get_state(transition.to).clone(),
+                    transition.trigger.clone(),
+                ));
+            }
+        }
+        transitions
+    }
```
bpandreotti: Instead of allocating a `Vec`, you can return an iterator:
```suggestion
    pub fn get_all_transitions(&self) -> impl Iterator<Item = (State, State, Trigger)> + use<'_> {
        self.all_states.iter().flat_map(|state| {
            state.transitions.iter().map(|transition| {
                (
                    state.clone(),
                    self.get_state(transition.to).clone(),
                    transition.trigger.clone(),
                )
            })
        })
    }
```

This `impl Iterator<...> + use<'_>` syntax is a bit cursed, but it basically means that this function returns an unnamed type that implements `Iterator`, and that uses the anonymous lifetime of the reference `&self`.

This is a very minor optimization, so feel free to ignore it.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+        for state in &mut new_states {
+            state.accept = !state.accept;
+        }
+
+        Automaton {
+            name: format!("{}_complement", self.name),
+            all_states: new_states,
+            initial_state: self.initial_state,
+        }
+    }
```
bpandreotti: Two notes:
1. This only works on automata that are already complete, right? I see you already call `determinize` when you use this, but you should document this assumption in this method.
2. Why are there two implementations of complement, here and in `operations.rs`? The second one is currently unused, but I assume it will be used in later PRs. If possible, it would be nice to merge them.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+                    })
+                    .collect();
+            }
+            new_states
+        }
+
+        fn rec_create_from_regex_operators(
+            pool: &mut dyn TermPool,
+            t: &Rc<Term>,
+        ) -> Result<Automaton, CheckerError> {
```
bpandreotti: No need to have a inner recursive function with the exact same signature. You should just make `create_from_regex_operators` itself recursive.

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+        next.dedup();
+        next
+    }
+
+    /// Checks membership by simulating the NFA directly: the set of reachable
+    /// states (kept as a sorted list, so its size tracks the automaton's
+    /// actual nondeterminism rather than its state count) is advanced per
+    /// input character and epsilon-closed using per-state closures computed
+    /// once. This avoids the potentially exponential subset construction of
+    /// determinization.
```
bpandreotti, line 841: Instead of keeping the current states as a sorted `Vec`, you can use a `BTreeSet`. I'm not sure what you mean by "track[ing] the automaton's actual nondeterminism rather than its state count", since you call `sort_unstable` and `dedup` in `step`, so it should be equivalent to using a `BTreeSet`.

carcara/src/ast/evaluate.rs
```diff
@@ -20,6 +20,7 @@ impl Value {
             Constant::Integer(i) => Value::Integer(i),
             Constant::Real(r) => Value::Real(r),
             Constant::String(s) => Value::String(s),
+            Constant::RegLan(_, _) => panic!("regular language constants cannot be evaluated"),
```
bpandreotti: Panicking here is a problem.

It's fine, of course, to not have support for evaluating some terms (as seen by the many "TODO"s in this file hehe), but we should fail gracefully instead of panicking. If the proof has an `evaluate` step that uses `RegLan` constants inside it, we should at most reject it and give an error, but never panic.

In `eval_op`, for example, we return `None` for unsupported cases, which gets handled later in `evaluate_impl`.

There are two ways to fix this, and I think they are both pretty easy:
1. Make `Value::from_constant` return `Option<Self>`, and `None` in the case of regular languages. Then, `evaluate_impl` should handle this case like it does for `eval_op`.
2. Add a representation for regular language constants in `Value`. This can be just the `(String, Automaton)` pair that is used in `Constant::RegLan` (I think it would behave as expected with equality and hashing, but you should make sure).

I think (1) would be simpler, but (2) gives us the advantage that terms like `(= (re.from_automaton ...) (re.from_automaton ...))` would evaluate correctly. What do you think?

bpandreotti (changes requested):
Some more detailed comments, but there's still some parts I didn't review.

carcara/src/automata/mod.rs
bpandreotti: Ah, actually you can't shadow, since you are constructing the `HashSet` iteratively. You could if you did it like this, though:
```rust
        let accepting_states: HashSet<_> = accepting_states.into_iter().collect();
```

carcara/src/automata/mod.rs
vinisilvag: Fair point! OSTRICH relies on the `dk.brics.automaton` Java library, where transitions/alphabets are defined using 16-bit Java chars. By using `u16`, we keep the data representation 1:1 with the solver's automata format, which also uses this range for all automata in the proofs. Considering scalability and potential future extensions, I'm not sure this is the best choice, but for now, it seems reasonable.

carcara/src/automata/mod.rs
vinisilvag: I wasn't familiar with `IndexSet`, but after looking into both alternatives and considering the operations we perform and the access trade-offs, I'm also convinced that `BTreeSet` is a good choice here. In our case, having two automata with the same transitions produce different hashes depending on insertion order doesn't seem desirable, especially given the comparisons we'll be performing. This makes `BTreeSet` an even better fit. We can go with it!

carcara/src/automata/mod.rs
bpandreotti: Ah, that makes sense. Fair enough then, I say we keep using `u16` and if someone ever needs an automaton over emoji or something weird like that we can move to `u32`.

vinisilvag: Actually, after discussing this with @HanielB  today, we came to the conclusion that even though OSTRICH generates proofs using transitions with `u16`-sized ranges, the operations themselves would not be incorrect if we used `u32`. In fact, if a proof were ever generated using the full Unicode alphabet, using `u32` would provide broader support. So I think we can go with `u32`, as you suggested!

carcara/src/automata/mod.rs
vinisilvag: This is cool, I'll implement it! I agree that avoiding the `Vec` allocation would be nice, but out of curiosity, is it likely to be that costly in this particular scenario?

carcara/src/automata/mod.rs
bpandreotti: Okay, sounds good!

carcara/src/automata/mod.rs
bpandreotti: It probably wouldn't be that bad. I think it's about equivalent to cloning the Automaton an extra time, so not that bad unless it's used in a critical loop.

vinisilvag: Makes sense!

carcara/src/automata/mod.rs
vinisilvag: That was exactly the case. The unused version will be needed in future PRs, but I agree that it would be better to keep just one of them. I'm thinking about merging them by checking whether the automaton is totalized (a complete DFA returned by the `determinize` function should be fine, I guess) and, if not, totalizing it before flipping all accepting states. I just don't know where the best place would be to keep this function. Do you think it makes more sense to keep it in `mod.rs` or `operations.rs`?

carcara/src/ast/evaluate.rs
vinisilvag: As far as I understand, (2) would only give us syntactic equality. However, for automata, we are also interested in semantic equality: two automata `A_1` and `A_2` should also be considered equal if they recognize the same regular language, i.e. `L(A_1) = L(A_2)`. That's why we implemented `is_equivalent`, which we will use in the verification procedures for this calculus.

In that sense, I don't think (2) would be correct for our use case, since using the `(String, Automaton)` representation would make equality depend on the representation of the automata rather than on the language they recognize. We would also have to be careful with `Hash`, since Rust requires that `a1 == a2` implies `hash(a1) == hash(a2)`, which would be difficult to guarantee for semantic equality (correct me if I'm wrong, please).

(2) seems like a much better idea to me, but given these considerations, (1) seems like the safer option. What do you think?

carcara/src/automata/mod.rs
```diff
@@ -0,0 +1,1372 @@
...
+/// A transition leads from the enclosing source state to a destination state
+/// and is labeled by a trigger describing which input symbols enable it.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd)]
+pub struct Transition {
+    /// Identifier of the destination state.
+    to: StateId,
+
+    /// Label of the transition, describing the consumed input.
+    trigger: Trigger,
+}
```
bpandreotti: Now clippy is warning that the `PartialOrd` and `Ord` implementation must agree. The easiest way to do this is to implement `PartialOrd` using our existing `Ord` implementation:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct Transition {
    /// Identifier of the destination state.
    to: StateId,

    /// Label of the transition, describing the consumed input.
    trigger: Trigger,
}

impl PartialOrd for Transition {
    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
        Some(self.cmp(other))
    }
}
```

carcara/src/ast/printer.rs
```diff
@@ -543,6 +543,7 @@ impl fmt::Display for Constant {
                 }
             }
             Constant::String(s) => write!(f, "\"{}\"", escape_string(s)),
+            Constant::RegLan(_, a) => write!(f, "{:?}", a), // TODO: comeback to this
```
bpandreotti: If I understood the code correctly, the string associated with the `RegLan` variant is the original string given to `re.from_automaton`, right? If so, you can just reconstruct the original term when printing:
```suggestion
            Constant::RegLan(s, _) => write!(f, "(re.from_automaton \"{}\")", s),
```
In any case, you should document the `Constant::RegLan` variant saying what the associated string is.

carcara/src/automata/mod.rs
bpandreotti: I think it's better to put it in `operations.rs`

carcara/src/ast/evaluate.rs
bpandreotti: Cool, that makes sense. Let's go with option (1) then.

carcara/src/automata/mod.rs
vinisilvag: By that, we meant contrasting a dynamic set of active states (representing the possible paths while processing the input, whose size is typically tiny) with a static bitset/array whose size scales with the total number of states (`|Q|`). I'll try to make this clearer in the comment.

Anyway, I like the idea of using `BTreeSet` here, so we can go with that!

carcara/src/ast/printer.rs
vinisilvag: Okay!

carcara/src/automata/mod.rs
```diff
@@ -244,18 +247,16 @@ impl Automaton {
     }
 
     /// Returns a vector of all transitions as `(source_state, destination_state, trigger)` triples.
```
bpandreotti: This comment is now outdated. Change "a vector of ..." to "an iterator over ...".

PR #139, opened 2026-08-24
Check rare hole rules

caotic123:


HanielB:
This PR is way too big. Break it down in smaller parts. Also remember to add descriptions to pull requests.

PR #140, opened 2026-08-24
DRUP resolution - Elaboration

caotic123:


carcara/src/elaborator/drup.rs
```diff
@@ -0,0 +1,185 @@
...
+    let mut ids = IdHelper::new(&step.id);
+
+    for (action, argument) in trace.into_iter().zip(&step.args) {
+        let DRupProofAction::RupStory(goal_literals, history) = action else {
+            continue;
+        };
+        let goal_clause = match_term!((cl ...) = argument)
+            .ok_or(ElaborationError::InvalidDrupTrace(
+                "addition argument is not a clause",
+            ))?
```
bpandreotti, line 121: We can use `match_term_err!` here instead of needing a dedicated error:
```suggestion
        let goal_clause = match_term_err!((cl ...) = argument)?.to_vec();
```

carcara/src/elaborator/drup.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    let mut weakened_clause = current_clause.to_vec();
+    for term in &goal_clause {
+        let literal = term.remove_all_negations_with_polarity();
+        let literal = (literal.0, literal.1.clone());
+        if !current_literals.contains(&literal) {
+            weakened_clause.push(term.clone());
+        }
+    }
```
bpandreotti, line 68: Here, when we convert a term to a `Literal` using `remove_all_negations_with_polarity`, we are losing information about double negations (i.e., `a` and `(not (not a))` are considered the same). So, if `goal_clause` contains `(not (not a))` and `current_literals` contains `a`, or vice-versa, we might falsely consider these terms as identical, and produce an invalid `reordering` step. This minimal example encounters this issue:
```
(assume a0 (not (not a)))
(step t1 (cl a) :rule drup :premises (a0) :args ((cl a)))
(step t2 (cl) :rule hole)
```

When elaborated, this produces:
```
(assume a0 (not (not a)))
(step t1.t1 (cl a) :rule reordering :premises (a0))
(step t2 (cl) :rule hole)
```
...which is invalid. To reproduce this, note that you must disable the `reordering` elaboration pass, otherwise it will remove the reordering step (i.e., use `--pipeline local`).

I'm not sure what's the best way to address this. One way would be to add a `not_not` step for each of these literals, and apply resolution with the `reordering` step. Since this is a very weird edge case, I think it's also acceptable to just not have support for this: maybe just detect this edge case and panic or return an error.

bpandreotti (changes requested):

carcara/src/elaborator/drup.rs
HanielB: Does the conversion to a literal need to remove all negations? Not doing that, and adapting everything else to not rely on this, would be the principled solution, no?

bpandreotti: If we just removed one, I think we would be considering `(not (not a))` and `(not a)` two unrelated literals, instead of the same literal with two polarities. Might be tricky to handle.

Maybe what would work instead is to use `remove_all_negations`, which returns a `(u32, &Rc<Term>)` with the number of leading negations removed.

HanielB: > Maybe what would work instead is to use remove_all_negations, which returns a (u32, &Rc<Term>) with the number of leading negations removed.

I didn't get how this would help.

bpandreotti: With that representation, `a` and `(not (not a))` would be different literals (`(0, a)` and `(2, a)`, respectively), and the presence of one in `current_literals` would not cause the second one to be skipped in the
```rust
if !current_literals.contains(&literal) {
```
line. Since currently they are both represented with `(true, a)`, we have that collision, which is what I think is causing the bug.

HanielB: I see. Yes, these are different literals, so that sounds right.

PR #141, opened 2026-08-24
Alethe2Eunoia: invocation from CLI

Mallku2:
This PR focuses on modifications to the CLI, to be able to invoke the compilation routines to Eunoia:
- Modifies the `cli/src/main.rs` `cli/src/app.rs` files, adding the corresponding structures to represent the input arguments when invoking carcara, the parsing of them and the invocation of the corresponding compilation routines.
- Modifies `translation/eunoia/alethe_signature/theory.rs` abstraction layer, to be able to build the path to the eunoia signature files, as passed by the user through the CLI, and modifies the AST and printer, to include the Eunoia command `include`.
- Misc.: minor fixes in `translation/mod.rs`, to fix an issue introduced on a previous PR, when dealing with the translation of Alethe `anchor` commands that do not introduce assumptions.

NOTE: could it be that GitHub is incorrectly comparing the branch `translation-to-main` with an older version of branch `main`? It is giving me a `can't automatically merge` warning, but the compared files do not match current `main` version...


bpandreotti:
Yeah, I think GitHub got confused after we merged the previous PR to main. What I would do is to rebase the latest commits from this branch (since the last commit that was merged, namely 8c5f00a5) on top of main, using:
```bash
git switch translation-to-main && git rebase --onto main 8c5f00a5
```
The resulting history looks like this:
```
48a9f68b (HEAD -> translation-to-main) mod.rs: removing tstp module export
d9ef8665 printer: fixing concrete syntax of include commands; fixed minor bug when translating empty anchors
39db6cfa cli: added translate command; compiler to eunoia: several changes in order to accomodate changes in cli
8d37b8a1 (origin/main, origin/HEAD, main) Alethe2Eunoia and Alethe2Tstp compilation routines (#135)
6624ea80 Update link to the specification.
...
```
This will require force-pushing, and we would lose the history of the `translation-to-main` branch from before the merge, but I think it's cleanest way to solve this.

Mallku2:
> Yeah, I think GitHub got confused after we merged the previous PR to main. What I would do is to rebase the latest commits from this branch (since the last commit that was merged, namely [8c5f00a](https://github.com/ufmg-smite/carcara/commit/8c5f00a5ad07088662da54e7a7838a09b93e5633)) on top of main, using:
> 
> ```shell
> git switch translation-to-main && git rebase --onto main 8c5f00a5
> ```
> 
> The resulting history looks like this:
> 
> ```
> 48a9f68b (HEAD -> translation-to-main) mod.rs: removing tstp module export
> d9ef8665 printer: fixing concrete syntax of include commands; fixed minor bug when translating empty anchors
> 39db6cfa cli: added translate command; compiler to eunoia: several changes in order to accomodate changes in cli
> 8d37b8a1 (origin/main, origin/HEAD, main) Alethe2Eunoia and Alethe2Tstp compilation routines (#135)
> 6624ea80 Update link to the specification.
> ...
> ```
> 
> This will require force-pushing, and we would lose the history of the `translation-to-main` branch from before the merge, but I think it's cleanest way to solve this.

It looks like it is fixed now. Yesterday, before opening the PR, I did merge with `main`, though I didn't push again, to commit the changes made during merge. For some reason that made github to compare the content of this PR against older versions of the modified files, that were not even the correct version from the current `main` version. It is now ready for review.

cli/src/main.rs
```diff
@@ -307,3 +309,55 @@ fn generate_lia_problems_command(options: ParseCommandOptions, use_sharing: bool
...
+        rules.as_deref(),
+        options.parsing.into_config(),
+    )?;
+
+    // NOTE: currently supporting only translation into Eunoia.
+    match &options.target {
+        TranslationTarget::Eunoia => translate_2_eunoia_command(
+            &alethe_problem,
+            &mut alethe_proof,
+            &options.eunoia_mech.unwrap(),
```
bpandreotti: Here, we are panicking if the user does not provide the `--eunoia-mech` argument (try running `cargo run -- translate eunoia /dev/null /dev/null`). Instead, you should make that flag required, by changing the `eunoia_mech` field in `TranslateCommandOptions` to be just a `String`. Later, when we implement other translation targets, we can have a more clever solution.

bpandreotti (changes requested):

cli/src/main.rs
Mallku2: You are right! Done.

bpandreotti (approved):

Issue #91, opened 2025-06-11
Improve how operator typing rules are specified and checked


bpandreotti:
Currently, the type checking for all operator terms happens inside the `Parser::make_op` function, in a huge (and ever increasing) `match` statement. This means that to add a new operator, a maintainer needs to modify this function in the parser, and implement the type checking rules manually, which can be non-trivial, and has lead to bugs before. It would be better if this type checking logic were in a different module, and if there was some simpler way to specify the typing rules for an operator.

The ideal version of this would be if a maintainer could add a new operator without having to even touch rust code, by just editing a configuration file. Carcara could then read this file at compile-time, and enforce the type checking rules specified in it. One way this might look is a `.toml` file, like this:
```toml
[[operator]]
name = "str.replace_re"
return_type = "String"
arg_types = ["String", "RegLan", "String"]
```

Alternatively, we can use the syntax used by [SMT-LIB theory declarations](https://smt-lib.org/theories-UnicodeStrings.shtml):
```
(str.replace_re String RegLan String String) 
```

It might be the case that not all operators can be succintly defined with this configuration language, especially considering the behaviour of some operators changes depending on parser config. In that case, I think it makes sense to have built-in operators, whose typing rules are still defined in rust, as well as the custom operators:

```rust
enum Operator {
    Add,
    Sub,
    // ...
    Custom(CustomOperator), // Read from config file
}
```

caotic123:
No @HanielB, just mapping, for example, a function that maps` (str.replace_str)` to `FunctionSort String (FunctionSort String String)`. We just need to add Variadic types, for example, `or` could be mapped to `VariadicFunctionSort String`. So now we test types, not their terms.

PR #143, opened 2026-09-04
Operator type-checking rules

netolcc06:
### Deferred operator type-checking rules (#91)

Splits the monolithic type-checking `match` in `Parser::make_op` into `src/parser/operators.rs`,
and adds an opt-in TOML config file (`custom_operators.toml`, read at compile time by `build.rs`)
for declaring new operators without touching Rust.

**What changed:**
- `operators.rs`: a small declarative rule DSL (arity + arg-sort constraints) replaces the
  ~400-line hand-written match for the ~150 built-in operators. Only ~10 operators (dynamic/
  config-dependent behavior, e.g. `RealDiv` literal folding, `Select`/`Store`, `ReFromAutomaton`)
  still need hand-written logic; everything else is a one-line rule.
- `Operator::Custom(CustomOperator)`: a new variant for operators declared in
  `custom_operators.toml` (`name` / `return_type` / `arg_types`). `build.rs` parses the file and
  codegens a static table; unknown sorts, duplicate names, or malformed entries fail the build.
- Shipped with 4 real cvc5 string extensions (`str.rev`, `str.to_lower`, `str.to_upper`,
  `str.update`) that carcara didn't implement as built-ins, as a working example.

**Open questions for discussion:**
- The TOML schema only supports the 5 primitive sorts (`Bool`/`Int`/`Real`/`String`/`RegLan`) -
  no `BitVec`/`Array`/`Set`/`Tuple`, since those need extra parameters or recursion. Worth it to
  extend, or keep custom operators intentionally simple?
- Custom operators are purely uninterpreted (no evaluation/rewrite semantics, no checker rules
  reference them) - is that the right ceiling for this feature, or should there be a path to give
  them real semantics later?
- `custom_operators.toml` lives at the repo root and is baked in at compile time (not runtime-
  configurable per invocation). Is compile-time the right granularity, or should downstream users
  be able to point at their own file without forking?
- No general TOML parser was added as a dependency - `build.rs` hand-parses this restricted
  schema. Fine for now, but would block extending the schema much further.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,146 @@
+//! Reads `custom_operators.toml` and generates `$OUT_DIR/custom_operators.rs` at build time.
+
+use std::fmt::Write as _;
+
+/// One `[[operator]]` entry parsed from the config file.
+struct OperatorEntry {
+    name: String,
+    return_type: String,
+    arg_types: Vec<String>,
+}
+
+/// Maps a supported sort name to its `Sort` variant token, or `None` if unsupported.
+fn sort_token(name: &str) -> Option<&'static str> {
+    match name {
+        "Bool" => Some("Sort::Bool"),
+        "Int" => Some("Sort::Int"),
+        "Real" => Some("Sort::Real"),
+        "String" => Some("Sort::String"),
+        "RegLan" => Some("Sort::RegLan"),
+        _ => None,
+    }
+}
+
+/// Strips a `"..."` quoted string, panicking with `context` on malformed input.
+fn parse_quoted_string(value: &str, context: &str) -> String {
+    let value = value.trim();
+    let inner = value
+        .strip_prefix('"')
+        .and_then(|s| s.strip_suffix('"'))
+        .unwrap_or_else(|| panic!("{context}: expected a quoted string, got `{value}`"));
+    inner.to_string()
+}
+
+/// Parses a `["a", "b", ...]` list of quoted strings, panicking with `context` on malformed input.
+fn parse_string_list(value: &str, context: &str) -> Vec<String> {
+    let value = value.trim();
+    let inner = value
+        .strip_prefix('[')
+        .and_then(|s| s.strip_suffix(']'))
+        .unwrap_or_else(|| panic!("{context}: expected a `[...]` list, got `{value}`"));
+    let inner = inner.trim();
+    if inner.is_empty() {
+        return Vec::new();
+    }
+    inner
+        .split(',')
+        .map(|s| parse_quoted_string(s, context))
+        .collect()
+}
+
+/// Parses one `[[operator]]` chunk (the text following the `[[operator]]` marker, up to the next
+/// one) into an `OperatorEntry`.
+fn parse_operator_chunk(chunk: &str) -> OperatorEntry {
+    let mut name = None;
+    let mut return_type = None;
+    let mut arg_types = None;
+
+    for line in chunk.lines() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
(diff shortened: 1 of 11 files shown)


build.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    let mut entries = Vec::new();
+    let mut seen_names = std::collections::HashSet::new();
+    for chunk in chunks {
+        let entry = parse_operator_chunk(chunk);
+        if !seen_names.insert(entry.name.clone()) {
+            panic!("duplicate custom operator name `{}`", entry.name);
+        }
+        entries.push(entry);
+    }
```
bpandreotti, line 112: Instead of manually implementing a TOML parser, it might be better to just use the [`toml`](https://crates.io/crates/toml) crate.

src/ast/custom_operator.rs
```diff
@@ -0,0 +1,43 @@
...
+pub fn lookup_operator(name: &str) -> Option<super::Operator> {
+    use std::str::FromStr;
+    if let Ok(op) = super::Operator::from_str(name) {
+        return Some(op);
+    }
+    CUSTOM_OPERATORS
+        .iter()
+        .find(|def| def.name == name)
+        .map(|def| super::Operator::Custom(CustomOperator(def)))
+}
```
bpandreotti, line 29: Instead of generating an array of `CustomOperatorDef`s, we could generate a custom operator enum with the variants specified in the TOML file, and a method that looks up the operator variant by name. The generated code might look something like this:
```rust
enum CustomOperator {
    Foo,
    Bar,
}

impl CustomOperator {
    fn lookup(name: &str) -> Self {
        match name {
            "foo" => CustomOperator::Foo,
            "bar" => CustomOperator::Bar,
            _ => unreachable!(),
        }
    }

    fn def(&self) -> &'static CustomOperatorDef {
        match self {
            CustomOperator::Foo => &CustomOperatorDef { name: "foo", ..todo!() },
            CustomOperator::Bar => &CustomOperatorDef { name: "bar", ..todo!() },
        }
    }
}
```
What do you think?

build.rs
bpandreotti, line 1: This solution works great, but I would prefer to implement this as a procedural macro instead of a build script. That is, a macro that when called as `custom_operators!("custom_operators.toml")` would read the toml file, and produce the generated code. Since we already have to `include!` the generated file anyway, this would just simplify things.

bpandreotti (commented):
The overall idea looks great, I only have a few suggestions on how to structure things.

build.rs
netolcc06: Switched to toml/serde for parsing.

src/ast/custom_operator.rs
netolcc06: Implemented close to this, with two differences: lookup returns `Option<Self>` instead of panicking via `unreachable!()` on an unknown name, since callers need to tell "not a custom operator" apart from a bug - an arbitrary symbol just not being any operator (custom or built-in) is the common case. Let me know what you think.

netolcc06:
Went through cvc5's documented non-standard/extended theories to see what else could go in `custom_operators.toml` under the current schema (`Bool/Int/Real/String/RegLan` only). Strings and Transcendentals are the only ones that fit that constraint, and carcara already covers everything in both except `str.rev`, `str.to_lower`, `str.to_upper`, str.update, which I added as example. Non-standard (Bags, Sequences, Datatypes, Separation Logic, most of Sets/Relations) needs a sort the schema doesn't support (Bag, Seq, user datatypes, heap/ref sorts, Set/Tuple-of-arbitrary-element). Should I try to extend the schema to cover those, or is 5 primitive sorts the right ceiling for "custom" operators for now?

PR #144, opened 2026-09-05
Add RARE CLI and parser configuration

caotic123:


src/bin/cli/app.rs
```diff
@@ -198,6 +198,22 @@ pub struct CheckingOptions {
...
+    #[clap(long, requires = "check-hole-rewrites")]
+    pub print_egglog: bool,
+
+    /// Keep running RARE egglog schedules until goals are saturated.
+    #[clap(long = "continuous-saturation", requires = "check-hole-rewrites")]
+    pub continuous_saturation: bool,
+
+    /// Cooperative time budget in milliseconds for each RARE hole rewrite.
+    #[clap(long, requires = "check-hole-rewrites", value_name = "MILLISECONDS")]
+    pub rare_check_timeout: Option<u64>,
```
bpandreotti, line 216: These `requires` values are wrong, they should be `"check_hole_rewrites"`, with underscores. The current version panics when running without `--release`, try `cargo run -- check --help`.

src/parser/rare/mod.rs
```diff
@@ -1,125 +1,166 @@
...
+        // Accept both `T` and `@T` references for RARE type parameters declared as `Type`.
+        if matches!(sort.as_ref(), Sort::Type) {
+            let alias = if let Some(stripped) = name.strip_prefix('@') {
+                stripped.to_owned()
+            } else {
+                format!("@{name}")
+            };
+            self.declare_symbol(alias, sort.clone());
+        }
+
```
bpandreotti, line 56: I had changed this so this behaviour is controlled by a [flag in the parser config](https://github.com/ufmg-smite/carcara/blob/126c53314131a694fea8ee9069981a5a639a04d0/src/parser/mod.rs#L106-L110), so we shouldn't need this extra declaration anymore.

I thought the flag was a better way to handle it, but if you prefer to have the extra declaration here, feel free to remove the flag instead.

src/parser/mod.rs
```diff
@@ -1302,12 +1302,13 @@ impl<'p, 's> Parser<'p, 's> {
             self.next_token()?;
             self.expect_token(Token::OpenParen)?;
 
-            // If the rule is `hole` and `--parse-hole-args` is not enabled, we want to allow any
-            // invalid arguments, so we read the rest of the `:args` attribute without trying to
-            // parse anything
```
bpandreotti, line 1307: Instead of simply removing this comment, it would be nice to replace it with an explanation of the new behaviour (which is, from what I understand, keeping the first string argument even if `parse_hole_args` is false).

bpandreotti (changes requested):

PR #145, opened 2026-09-20
Elaboration of `distinct_elim`

HanielB:
The `distinct_elim` checker accepts each pairwise disequality in either orientation, but the rule specifies only `(not (= aᵢ aⱼ))` with `i < j`. Steps using the other orientation are now rewritten into a `trans` of a `distinct_elim` step with the correct elimination and a polyeq bridge back to the original conclusion.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


src/elaborator/polyeq/clausification.rs
```diff
@@ -1,14 +1,94 @@
...
 
+fn is_flipped(term: &Rc<Term>, a: &Rc<Term>, b: &Rc<Term>) -> bool {
+    let diseq_args = match_term!((not (= x y)) = term).unwrap();
+    diseq_args != (a, b)
+}
+
+/// The elimination `distinct_elim` specifies, obtained by traversing
+/// `got` alongside the pairs of `args` and rebuilding only the
+/// disequalities that are flipped. If there is nothing to flip back,
+/// so the step needs no elaboration.
```
bpandreotti: The last sentence in this comment is a bit confusing. Did you mean "returns None if there is nothing to flip back ..."?

bpandreotti (changes requested):

bpandreotti:
Another thing: you should also update the list of elaborated rules in `docs/src/elaboration/polyeq.md`, to add `distinct_elim`.

bpandreotti (approved):

